PR #18, opened 2017-05-29
Try having multiple s3 providers, one per OS

DDRBoxman:


Palakis:
Here is the bucket website for OS X builds : http://obs-ndi-osx-builds.s3-website.eu-central-1.amazonaws.com/

PR #229, opened 2018-09-27
Reduce active CPU usage by 80%, disconnected CPU usage by ~ 95%

### LINKED ISSUE #70: High CPU usage on Mac OS
I'm attempting to stream from one Mac to another via NDI and a crossover cable and everything is working very well with the exception of high CPU usage on the sender.  Outputting at 1280x720@30 output seems to cause about a 30% CPU load, whereas 1920x1080@30 is around 70%.  This is more or less in line with what I see when using the x264 software encoder, whereas CPU usage while using the Apple VT h.264 hardware encoder is somewhere around 3%.  My guess is that the NDI output is not taking advantage of the hardware encoder.  Is there a way to specify how the output is encoded or am I stuck with software only?  For reference, I'm using OBS 20.1.0/OSX 10.11 on both ends.

fredemmott:
- receiving returns immediately if not connected, meaning the poll threads are busy-looping. Add a 100ms sleep in these threads if disconnected to avoid this
- profiling shows that libndi has a mutex when receiving frames - so trying to pull frames in two threads is slower than pulling in one. Move to one thread
- now that we're in one thread, we can block the poll for as long as we like - poll without a timeout of 100ms instead of 1ms

Results:
- When there is no NDI transmitter on the network: before: 2 cores at 100% + 1 at 80%  after: 1 core at 20% which is pretty much what OBS takes up without NDI
- When there is an active source: before 2 cores at 100% + 1 at 60%. after: 1 core at 60%

fixes #70

### DIFF
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -59,8 +59,7 @@ struct ndi_source
 	int sync_mode;
 	video_range_type yuv_range;
 	video_colorspace yuv_colorspace;
-	pthread_t video_thread;
-	pthread_t audio_thread;
+	pthread_t av_thread;
 	bool running;
 	NDIlib_tally_t tally;
 	bool alpha_filter_enabled;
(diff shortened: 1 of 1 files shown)


fredemmott:
FWIW, this lets a 2013 Mac Pro record 1080p30 at ultrafast; without this, I get 1080p2 or so :p

PR #299, opened 2019-03-31
RPM support and lib64 dir fix

### LINKED ISSUE #137: RPM Package
Is possible to create this package?

blackPantherOS:
Hi!

First pull provide Spec for RPM-build, related with this https://github.com/Palakis/obs-ndi/issues/137
The working rpm packages available our official repositories

Second patch fix the /usr/lib vs /usr/lib64 problem on 64 bit systems like Fedora/blackPanther/etc

Regards

Closes #137


obs-ndi.spec
```diff
@@ -0,0 +1,107 @@
+#*********************************************************************************************************
```
gbraad: :-/

obs-ndi.spec
```diff
@@ -0,0 +1,107 @@
...
+Name: 	 obs-ndi
+Version: 4.6.0
+Release: %mkrel 2
+Source0: %name-%version.tar.gz
+Source1: obs-studio-23.1.0-rc1.tar.gz
+Patch0:	 %name-libdir.patch
+Summary: OBS broadcast to OBS
+License: GPL
+Group:   Multimedia/Video
+Requires: %{_lib}ndi3 >= 3.8.0
```
gbraad: is there an RPM spec providing this?

gbraad:
I suggest to aim for a spec that can be included to RPMFusion; a non-free for `libndi` and one for `obs-ndi` as free. Especially, as currently it is not required to have `obs-studio` installed?

```
 Requires: %{_lib}ndi3 >= 3.8.0
```

obs-ndi.spec
```diff
@@ -0,0 +1,107 @@
...
+#*************************************************************************************(c)2002-2019********
+# 	Soec written by Charles K Barcza for blackPanther OS - www.blackpanther.hu
+
+%define genlibndi 1
+
+Name: 	 obs-ndi
+Version: 4.6.0
+Release: %mkrel 2
+Source0: %name-%version.tar.gz
+Source1: obs-studio-23.1.0-rc1.tar.gz
```
gbraad: For Fedora this is provided by:

https://download1.rpmfusion.org/free/fedora/releases/29/Everything/x86_64/os/repoview/obs-studio-devel.html

CMakeLists.txt
```diff
@@ -133,7 +133,7 @@ if(UNIX AND NOT APPLE)
 		obs-frontend-api)
 
 	install(TARGETS obs-ndi
-		LIBRARY DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/obs-plugins)
+		LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}/obs-plugins)
```
gbraad: 👍

obs-ndi.spec
blackPantherOS: RPM generated with two different license originally, if user want can build from source, or use your deb package as well for generate a RPM package with non-free license. But you right! The "requires" my fold, I've fixed to as recommends optional package and fixed the license as well on my git

blackPantherOS:
We don't understand well what do you want but we have fixed the spec file.

This sample works well on blackPanther OS but we didn't test it on Fedora, please test it:

https://github.com/blackPantherOS/obs-ndi/blob/master/obs-ndi.spec

gbraad:
It also requires `avahi-daemon` to allow source announcement and disccovery via mDNS to work.

vwbusguy:
I was able to get an RPM built for Fedora but it ended up being a good amount of work from where it is now.  I managed to update everything to 4.8.0 as well and it loads.  I haven't tested it against actual hardware yet.  
This spec obvious needs some cleanup, especially as a number of things are hard-coded that shouldn't be, but it at least builds on Fedora 32.

https://oc.scottawilliams.com/nextcloud/index.php/s/4G99TkPcoz3qdMa

gbraad:
@vwbusguy I noticed you add to the changelog... can you create a new PR with the improvements, as the OP didn;t reply for almost a year

vwbusguy:
(Good to see you you here, Gerard)

Yeah, what I posted here was just how far I got to get it to compile.  It is definitely going to need some clean up before it's PR worthy, but I'll see if I can do this soon.

vwbusguy:
Also, the conflict is with cmakelists.txt.  It's been refactored since the original PR, but it's just two lines that need updated.

vwbusguy:
The people I was working with to build this for ended up needing NDIHDX drivers, so I don't actually have hardware to test this on.  I can see that everything loads as expected in OBS on Fedora 32.  If you want, i can clean it up and submit a PR, but I won't be able to test that it actually works besides OBS loading the module successfully.  The biggest change is that I used obs-studio-devel instead of requiring the OBS source directly.

gbraad:
Ref #311 and #137 
I can test from NDI HDX with iOS if needed.

kuboosoft:
https://github.com/Palakis/obs-ndi/issues/311

kwizart:
FYI, the obs-ndi, libndi and ndi-sdk packages are under review within RPM Fusion
Theses will land in the default distribution for Fedora 32 and later once reviewed.

paulpv:
All, I'll look in to this after I get 4.11.0 released

paulpv:
@blackPantherOS et all.
I have watched this PR and even commented a long time ago on 2023/02/28, but I knew that this plugin was or should be and now is based on https://github.com/obsproject/obs-plugintemplate and I wanted them to fix the problem.

At a minimum, my recent updates to obs-ndi's plugintemplate code makes the change to `CMakeLists.txt` unnecessary.
They already have that change in https://github.com/obsproject/obs-plugintemplate/blob/c9ba8eeb7c90ecaafe33c7ac6c970f6d4ac6fe82/cmake/linux/helpers.cmake#L30

Also, just today they pushed a PR that fixed installing the .deb on Ubuntu and that I confirmed works.

If you really want RPM support, are you able to create a **generic** PR of the `obs-ndi.spec` (again, made generic for other plugins) changes for the https://github.com/obsproject/obs-plugintemplate project?
If no, then I could give it a try, but it is low priority for me until after I get 4.12.0 released.

If they reject the PR then I could be willing to apply this patch to obs-ndi only.
If they accept the PR then I would obviously merge that change into obs-ndi.

PR #459, opened 2020-04-28
NDI Source: remove "Internal" from the options of the "Sync" property

Palakis:
That option never worked. It's just a very bad idea. This PR completely removes the "Internal" mode from the options of the "Sync" property on NDI Sources.

It also also sets "Source Timing" as the default value of the "Sync" property. Testing with both OBS and vMix as NDI senders show that this is the best option and allows the receiving OBS to "honor" the A/V timestamps of the source.


sch00lb0yq:
Internal did work for me just by connecting my 2 pc's with an ethernet cable

PR #482, opened 2020-05-09
source: add a switch to disable audio

norihiro:
This pull request adds a new feature to disable arriving audio from the NDI source.

A new checkbox 'Enable audio' is added to the property dialog for an NDI source.
obs-ndi-switch-audio


Palakis:
@norihiro Can you please fix the merge conflict?

Palakis:
Also, what is your use case for this feature?

norihiro:
My use case is as below.
- Using NDI source to capture a camera on a wireless device such as iPhone. Audio from NDI is not used.
- Audio comes from another audio interface connected via USB.

The problem is that OBS Studio keeps adding buffering for the audio. There is a drift of sampling frequency, which causes more buffering.

Palakis (approved):

Issue #534, opened 2020-07-20
Audio de-sync issues with 4.9 and 4.8 (3 PC setup)


Xciles:
Issue: 
- Audio de-sync using either OBS-NDI 4.8 or 4.9
- Worse audio de-sync when using the 4.9 version

Setup:
```
PC - v (NDI)
  Server (Encoding -> Twitch)
PC - ^ (NDI)
```

Both PCs have 4 NDI sources: 2 Video (Main Display capture 1080p (rendered down from 1440p native), Webcam capture 720p) and 2 Audio (Desktop Audio and Microphone). All sources are sent to the server which does the rest and combines additional shared scenes. We get no (or almost no) rendering issues or other dropped frames using this setup.

Splitting everything into different sources should have, as other topics described, resulted into everything being synced properly, however for some reason, this is not the case using our setup.

All 3 PCs are rather beefy and (should) have no issues regarding CPU or GPU limitations and are connected with a proper router on a gigabit network. Both source PCs are using 10%-ish CPU and 15/20% GPU and are sending roughly 120/160mbits of data. The server receives roughly 250/350mbits of data, uses 20%-ish CPU and 25/30%-ish GPU. We connect to the server using remote desktop with hardware graphics adapter enabled for remote sessions (which should resolve other rendering issues), however connecting without Remote Desktop and using a monitor with mouse and keyboard does result in the same issues. The server also has no (or almost no) rendering issues or other dropped frames.

In the below scenarios we tried all different sync methods, Internal, Network and Source Timing, which still results into the audio being de-synced. Some scenarios de-sync more and faster than others and when using 4.9 this de-sync happens a lot faster and makes 4.9 not usable. 

What we tried so far (and resulted in audio being de-synced (after a while >1 hour and faster <1 hour)):

- 4 NDI Sources per PC and using different sync methods.
  - The above scenario, 4 sources: 2 vid, 2 audio.
  - We tried this setup using OBS 25.* and NDI 4.8 or 4.9.
  - In this scenario all sources have a different delay. OBS NDI 4.9 makes the problem worse and increases the delay at a faster rate.
- 2 NDI Sources per PC and using different sync methods.
  - The above scenario, using 2 sources instead of 4. We combined the audio-only sources into the corresponding video sources:
    - Desktop capture with desktop audio source.
    - Webcam with microphone source.
[...]

Issue #530, opened 2020-07-13
Audio output very delayad


jwetzell:
OS: Windows 10 build 18362
OBS: 25.0.8
obs-ndi: 4.0.8

The Audio from the "Main" NDI Output is quite delayed (almost a second)....other NDI Sources were tested (HX from iPhone, NDI Output from Wirecast) and the audio is correct. However, if I were to add Dedicated NDI Output (audio only) to something the audio from that is in time with the NDI Main outputs video.

jwetzell:
update:
Mac OS 10.15.5 with same OBS/obs-ndi version and I don't experience the audio delay

simpleman19jenkins:
I am having the same issues running the same versions as well, mine seems to be around 1400ms and I am able to work around it using a render delay plus negative audio offsets.  However, I don't think the delay is constant as it still goes out of sync occasionally after lining the audio up.

simpleman19:
I am having the same issues running the same versions as well, mine seems to be around 1400ms and I am able to work around it using a render delay plus negative audio offsets. However, I don't think the delay is constant as it still goes out of sync occasionally after lining the audio up.

Trouffman:
Closing this issue as it is old/not been updated/expired and most- likely fixed in more recent version.
Please upgrade to the latest versions (OBS / plugin / runtime) and open a new issue if this is still a problem.

Issue #498, opened 2020-05-19
Audio sync exists with 4.9.0 (how to reproduce)


loopy750:
Audio sync is still an issue with 4.9.0, and tested using a few different network cards. The issue can be easily reproduced with both Normal and Low latency modes, although it may be less severe with the latter.

Please see the videos for how to reproduce, and the files used (linked/attached).

Latency: Normal - https://streamable.com/5p56f8
Latency: Low - https://streamable.com/b0dbsc

Over time, the audio/video may slowly drift out of sync. However, we can cause this effect immediately using the "static" image.

https://i.ibb.co/KGbPyPP/Image4.jpg - This is the expected result: frame "1" matches beginning of the 1kHz tone, hence in sync.

https://i.ibb.co/2Nch2Z6/Image3.jpg - The result after the "static" image has been shown: 1kHz tone begins at frame "54", 7 frames out of sync, or 116ms.

"Stress test" image file: https://i.ibb.co/SQhKMKQ/400pc-Gaussian-50pc-Grey.png
60 FPS Test mp4: [60 FPS TEST v2_1 START_AT_1.mp4_13MB.zip](https://github.com/Palakis/obs-ndi/files/4650132/60.FPS.TEST.v2_1.START_AT_1.mp4_13MB.zip)

WiredUK:
I'd love to know if this is an NDI thing or an OBS thing. We're also seeing audio sync issues, though the low-latency mode appears to be a lot better. Even if I had a way to "reset" the sync, that would be enough for us, it's really starting to become a big problem.

Anonymous2:
Hello, I also found a lot of audio/video desync with 4.9.0 that I never had on previous versions

manuelnyc:
@WiredUK the problem is THIS palakis NDI plugin, when I use newtek tools I never have an issue, no drops no nothing, but when I use this palakis the problems, frame drops, out of sync, start to happen, sadly this is the only NDI plugin for OBS

WiredUK:
> @WiredUK the problem is THIS palakis NDI plugin, when I use newtek tools I never have an issue, no drops no nothing, but when I use this palakis the problems, frame drops, out of sync, start to happen, sadly this is the only NDI plugin for OBS

Not necessarily. It could still be OBS. We really need to understand this, it's really hurting the quality of our streams.

manuelnyc:
[...]

Issue #379, opened 2019-12-19
audio is desynced 


benniadioz:
The Audio Stutters are away since i did make a new ndi source for every audio and video source to my encoding pc and Disabling Device Timestamps on the desktop audio and setting NDI to Network sync on the encoding system.But now i have a new Problem , the stutters are absolutely away but my audio from the Game and my Voice isn´t synced to the Stream...Must i add a delay to the ndi sources of the encoding Pc? And which obs you start first? Gaming pc or encoding pc? Much Greets from Germany , i hop i can get an answer to make this problem go away....im slowly going crazy...

paulpv:
Is this still an issue in https://github.com/Palakis/obs-ndi/pull/849 's 4.11.0-RC?

PR #564, opened 2020-09-25
Change audio output to use synthetic timecode.

darthsteven:
I had audio/video sync issues trying to use the NDI plugin to feed OBS video and audio into NDI and then on to Zoom.

After a lot of fiddling around, I found that the culprit was OBS/this plugin. I could use the NDI monitor to see the video + audio output and although they would start in sync they would gradually drift apart, which was pretty useless!

It seems like lots of people have been experiencing similar things with the video/audio sync of the output:

* #375 
* #379
* #498 
* #530
* #534 

In the end, I made a single line change and now the sync is near-perfect.

This single line has a very long history, and rightly so!

1. In 95c1d2ce21bd9bd6ee27161b5b67344ec605c917 the timestamp on the video and audio was changed from using the OBS timestamp to letting NDI synthesize the timestamp.
2. Then in d8d3952173ad5538ae8f19117e24540843d19cdd it was changed back to using the OBS timestamp from the audio frame. Note that at this point there was a TODO to determine what was the best thing to do.
3. In 57ae3a3fbb3a603b430e5362831f478337bca3ef this TODO was removed as part of general code style refactor, maybe the comment wasn't actioned it was just bad style?

It's changed a little bit since then, but fundamentally it's the same line of code: use the OBS timestamp for video and audio.

So, history lesson over, back to the single line change.
I added some super simple logging to output the timestamp of the frames we are getting from OBS in the raw_video and raw_audio callbacks, and the timestamps were essentially the same for the first couple but began to drift fairly quickly, they were literally seconds apart after 15 minutes. Maybe there's an OBS bug around the timestamps that are sent along, since the audio data seemed fine, it was simply out of sync.

I'm totally new of OBS so don't understand the internals really at all, but as far as I can tell, the video output is performed by a thread that's waiting for frames from the render, and that it's clocked at the framerate of the video. Then it looks like the audio thread generally tries to keep up with the video rendering, once a video frame goes out the corresponding audio frames are sent out too.

Assuming my understanding is correct then we should simply rely on the OBS timestamp from the *video* only, and let NDI synthesize the audio timecode, since the NDI library will do its thing and the audio will get the timecode of the previously sent video frame.

In my limited testing, this works just fine, and is a decent workaround for this plugin to use.

Maybe this should be raised as an upstream issue in OBS too, since it looks like the timestamps coming out of OBS audio frames are plain wrong.

### DIFF
--- a/src/obs-ndi-output.cpp
+++ b/src/obs-ndi-output.cpp
@@ -321,7 +321,7 @@ void ndi_output_rawaudio(void* data, struct audio_data* frame)
 	}
 
 	audio_frame.p_data = o->audio_conv_buffer;
-	audio_frame.timecode = (int64_t)(frame->timestamp / 100);
+	audio_frame.timecode = NDIlib_send_timecode_synthesize;
 
 	ndiLib->send_send_audio_v3(o->ndi_sender, &audio_frame);
 }


darthsteven:
And...having made the commit already, I've just read the commit guidelines, and it's wrong, sorry!
I can re-do?

tt2468:
@darthsteven Just amend the commit with the fixes and force push the changes to your branch, and it will apply here.

tt2468:
Sounds to me like some weigh-in needs to be made by @Palakis and @derrod as to what the reason behind the changes are. Perhaps this would be best implemented as a toggle in the settings, since it seems that there may be benefits and downsides to both behaviors.

Palakis:
In general, A/V drift in obs-ndi is caused by OBS' audio sync mechanism.

OBS adds a short buffering delay to its internal audio engine if some sources provide late timestamps. This can happen with media sources, where the timestamp provided to OBS by the media source matches the file's playback rate and not necessarily realtime. And so, because of the internal buffer growing to accomodate late sources and _keep audio and video timestamps in sync_, the raw video and raw audio callbacks don't necessarily output related data (audio being delayed internally by OBS).

In a nutshell, with OBS, the audio and video timestamps are actually perfectly in sync, but they don't necessarily get provided in sync to obs-ndi. Which for OBS isn't an issue, because the encoder and/or the end-user's player (not exactly sure which one) uses the provided timestamps to play audio and video in sync.

But for NDI, the SDK/runtime supposedly does not do any kind of A/V sync based on timestamps, rather relying on the sending application to provide audio and video already synced (or at the lowest capture latency possible).

@darthsteven Your fix and explanation sound promising. I can't wait to test it on my end. Things I don't understand though:

- How does making NDI synthesize audio timecodes fixes the issue?
- Do audio and video stay synced in the NDI output even if OBS outputs "adding X milliseconds of audio buffering" messages in the logs?
- Did you test and witness the sync fix with the NDI monitor?

Palakis:
So the culprit is either the stuff mentioned in the lengthy explanation above, or just bad timestamps math.

darthsteven:
So some more general discussion before I answer those specific questions:

I added some naff debugging code to simply log the timestamp coming from OBS for video and audio, using the (`blog` function I think) and at the start of the output the timestamps were really close together, but they started drifting apart really quickly. After a few minutes, they were seconds apart.
As far as I understand it, if we were writing the raw data to a file, then this would probably be fine, as you'd hope that whatever file writing library we were using would handle it and then the player would match the timestamps back up and done. Because NDI is realtime, I think that's the issue we have.
I was doing most of my testing with a media source (a few specific videos) but have seen this happen with other sources too (camera no audio + aux audio device even).
I should note that the actually audio data wasn't drifting, it's the right data with the wrong timestamp. Super weird.

> * How does making NDI synthesize audio timecodes fixes the issue?

I'm not 100% but my assumption is that NDI is trying to reconcile the drift between the timecodes that its given. I'd assume that if one supplied audio and video frames where the timecode was only out by a tiny amount it would match them up internally and provide a clean, synced feed to consumers.
By getting NDI to synthesize the timecode, the audio is effectively timecoded to the video, but by NDI, not by OBS + maths.
OBS was sending the plugin the correct video and audio data at the right time before, but the timestamps on the frames looked wrong.

> * Do audio and video stay synced in the NDI output even if OBS outputs "adding X milliseconds of audio buffering" messages in the logs?

I'm not sure, I've seen that message a few times, but didn't take much notice of when I saw that / vs. when it worked. Sorry! If I'm understanding correctly the implication of OBS adding this buffering, there would be no recovering from that, since then the audio/video feed would forever be out of sync anyway?

> * Did you test and witness the sync fix with the NDI monitor?

I did! I absolutely did. You can turn on the visual audio meter and then use a [audio/video sync video](https://www.youtube.com/watch?v=TjAa0wOe5k4) to purely _see_ the sync.

I also used this in a live production, where I took the NDI output, and used the NDI virtual camera to feed it into Zoom and had no complaints about the lip-syncing. Previously it was awful!

loopy750:
Issue #498 mentioned in the first post is my one. So I decided to perform 10 test recordings of both **_4.9.0 "stable"_** vs **_4.9.0 "audio_frame.timecode change"_**, the _audio_frame.timecode_ variable change mentioned here. (Basically I cloned tag 4.9.0, made that one change within **obs-ndi-output.cpp**, compiled, and replaced the stable _**obs-ndi.dll**_ with this one.)

Using my 'static' image appears to speed up any sync issues that might otherwise take hours to occur, athough it's possible streaming for hours might still be the ultimate test... I'm not sure. I'm not going to 100% suggest the fix here is 'the' fix, but my test results do look promising.

[60 FPS Recording tests]

**v4.9.0** - Before 'flashing' (briefly displaying) my static image, the audio was anywhere from in sync with the video to up to 18 frames out, and up to 23 frames out of sync after displaying the static image.

**v4.9.0 w/ audio_frame.timecode changed** - Before 'flashing' the static image, the audio was anywhere from in sync with the video to up to 4 frames out, and after displaying the static image there was little to no impact...if anything it brought it closer to being in sync with the video.

So in conclusion this looks extremely promising and is a huge difference compared to the current 4.9.0 release, although it would be good to see some more tests - or, as mentioned, an option to toggle between the current and proposed option within the plugin itself, as I think this could finally/hopefully be the solution that has annoyed many users of this plugin.

darthsteven:
@loopy750 that sounds great, can I ask how you're doing your tests, i.e. what are you using to record the NDI output etc.?
If you record directly in OBS, I'm assuming your recording there is perfectly in sync?

loopy750:
@darthsteven Yes, I'm simply just recording directly with OBS's record feature and saving as mp4 files. The sync wasn't perfect every time, but was usually within 10-20ms, vs +300ms at times with the current release. So I wouldn't say it was perfect, but whether that's because of the way it records in OBS, or the NDI plugin, I'm not exactly sure, but I do know the change has made it much much better, so it's a very good find.

darthsteven:
Oh, wait, so you're saying that the change here in the NDI plugin changes how OBS's 'built in' recording plugin is working?

loopy750:
Negative. What I'm suggesting is there still seems to be some slight sync variation with the change, but we're talking only a few milliseconds instead of hundreds of milliseconds. Those are my findings anyway. If anyone else can confirm, that'd be good to see too.

darthsteven:
@loopy750 So, if you were to completely remove the NDI plugin from your OBS instance, and re-run your tests (nothing seems to need NDI?) then your recording is totally perfect?

I wonder if another test would be to record the NDI stream from NDI Studio Monitor, and compare that.

loopy750:
Oh ok, I'll try and clear it up.

The test is recording an mp4 from the "RECEIVE" OBS. One OBS instance is on one PC "TRANSMITTING" to the other PC that is receiving the NDI transmission and recording it.

Although you do bring up an interesting question regarding general recording in OBS. And yes, with OBS totally out of the picture and just recording a video file playing in the "Source", there is some sync variation from recording to recording. So that could suggest that your OBS code change is "perfect" and it's OBS that slightly varies. Though, to be more confident in my findings, I'd want to see someone else perform the same test/s.

darthsteven:
@loopy750 ah hah! That makes perfect sense now. Sorry, I've only been using one OBS, so yeah, I can see what you're testing now. Great.

It would be good to get to the bottom of why the timestamp coming out of OBS for video/audio is sometimes so far apart. The 'fix' I'm suggesting is effectively a workaround for that (albeit a seemingly very effective one.)

I'm not going to have time to write it for the next few weeks, but maybe a debug output plugin that simply writes the timestamps it gets to a file would shed a little light on this. They should stay perfectly in sync, but from my limited testing they were seconds out (and more so in the latest unreleased versions of OBS.)

Mega61:
How can we make this change effective without needing an NDI update?

tt2468:
@Mega61 The only way to use this change is to install the version that CI built of this PR.

darthsteven:
@Mega61 You can download the plugin from https://dev.azure.com/Palakis/obs-ndi/_build/results?buildId=1415&view=artifacts&type=publishedArtifacts and then pop the built plugin into the right place in your OBS install.

aquarat:
I recently did my first wholly NDI shoot with 6-video sources (BirdDog NDI encoders). Unfortunately, several of the video sources drifted during the 1-hour shoot, in some cases up to 2 seconds. Intent on not letting that happen again I bought an HDMI splitter, so I could send the same signal (ffmpeg test signal with timecode burnt in) to all my NDI encoders. I've been testing syncing with this setup and have replicated the drift I saw during my first NDI live shoot (with the latest obs-ndi release).

It looks like this might not be an OBS or NDI problem per se but partially an encoder issue. The BirdDog encoders don't have NTP daemons installed, so their clocks don't sync. "Source Timing" appears to use the attached device's timecode, "Network Timing" appears to use encoder timecode. If the encoders were synced with NTP the Network Timing option should produce _really_ good results.

Unfortunately my cameras aren't synced, so can't use source timing and network timing is also out due to lack of NTP. The best option is timestamping the frames as they arrive, which is what this commit appears to do.

And it looks like it works **really** nicely. It's been 40 minutes now and everything is still exactly in sync (30fps), provided sources are set to "Low Latency". I haven't tested audio yet.

Thanks @darthsteven :)

awrebels:
@darthsteven do you have an updated build for the plugin? Doesn't look like there is anything in the artifacts folder and releases isn't accessible.

tt2468:
He will have to force push the commit again in order to trigger another CI build. For some reason I can't do it myself.

darthsteven:
@awrebels they are building now: https://dev.azure.com/Palakis/obs-ndi/_build/results?buildId=1586&view=artifacts&pathAsName=false&type=publishedArtifacts so they should appear there in a few minutes.

aquarat:
I've found audio glitches/distorts in OBS with this change. I've reverted to an older version of OBS-NDI that supports "internal" sync (to deal with sync drift on video) and I'm using Audinate's Dante for audio. I think the problem is that the encoders aren't synced. For comparison, the Dante system uses PTP for syncing. I'm using BirdDog 4K Flex and Minis as encoders.

awrebels:
@aquarat so are you using OBS only for video or are you somehow sending OBS audio to the Dante virtual soundcard, and then to the encoder?

aquarat:
The first fully NDI shoot I did (6 NDI video sources), I ran multiple wireless mics into a recorder/mixer and then out to a dedicated audio only camera. That went into OBS via NDI.

After much testing I now seperately run audio from wireless mics into a DAW (Ardour) via Dante AVIO hardware and then route the DAW output to the OBS machine, also using Dante, specifically Dante Via. OBS sees the audio as a sound card (WDM or ASIO). Dante Via and Dante virtual sound card handle the routing and syncing. The system works well, but is a bit costly if you include all the hardware... but it gives me incredible flexibility, professional audio interfaces - and it's fun to experiment. You don't need Dante hardware to make it work; Dante Via will route audio to/from other sound interfaces. Critically, Dante uses Precision Time Protocol to sync devices with very high precision and I suspect that's why its reliable.

bvkrieken:
@darthsteven 
Unfortunately the 4.9.1 build found here: 

https://dev.azure.com/Palakis/obs-ndi/_build/results?buildId=1586&view=artifacts&pathAsName=false&type=publishedArtifacts

is giving me trouble with audio from NDI sources. The VU meters for the NDI sources are constantly flashing red and doing weird stuff :) It also prevents me from recording. When I start a recording I will receive an error after 2 seconds saying "An encoder error occured while recording". When I remove the NDI sources from my scene and then press "Start recording" everything is fine. So I guess this build has indeed introduced other issues that make the plugin unworkable.

Is there a way you could fix this new issue?

darthsteven:
So, as much as I love open source and want to help I used OBS briefly while trying to get a particular workflow going and issues like this NDI forced our team to pursue alternative approaches, which have since worked out mostly just fine.
I don't have the time or resources to fix any issues here, sorry.

(I do have time to comment though, obviously!)

I would say that I thought that the change I made was purely to do with the way that the NDI plugin exports the main output into the NDI library, and thus I'd be surprised if this change caused issues with the NDI sources, which afaik are handled by entirely different code. However, I will refer you to something I said in my initial post:

> I'm totally new of OBS so don't understand the internals really at all

Best of luck!

bvkrieken:
@darthsteven Thank you for your honesty! 

This fix really got my hopes up but I guess it's back to waiting for a real fix by @Palakis 
We would love to use OBS with this plugin in combination with the NDI camera's in our studio, but the audio/video desync problem has forced us to use Xsplit Broadcaster instead, a real shame.

Palakis (approved):

aquarat:
Just reiterating that I find the audio distorts severely with this change. It distorted with the last major release too, but only every few seconds. This distortion is continuous.

darthsteven:
@aquarat is your OBS outputting audio via NDI in your setup via the main OBS output?

aquarat:
@darthsteven Apologies, I should have been clearer. The distortion is in the monitor output (toggled in advanced audio properties and routed out the machine's speakers). This is a Windows laptop. I don't have the NDI output turned on currently.

aquarat:
@darthsteven Apologies, the distortion was coming from my test video source. Please ignore my last comment. The change seems to be working well.

darthsteven:
Yeah, I was going to say, I might be wrong, but the code change here should only affect the main NDI output of the OBS programme feed, not monitors etc.

aquarat:
Yeah, it was a bit confusing because the distortion changed between the last release and the build for this change but I think that was just an unrelated timing issue.

bvkrieken:
Does this mean that Darthsteven's fix has been added to the plugin? Where can I download this new plugin version?

jlamur:
@bvkrieken It looks like no release with this fix has been done yet. You'll have to build the plugin yourself.

bvkrieken:
@jlamur Ah ok, thank you for your answer :)

comemyfanatics:
> @Mega61 You can download the plugin from https://dev.azure.com/Palakis/obs-ndi/_build/results?buildId=1415&view=artifacts&type=publishedArtifacts and then pop the built plugin into the right place in your OBS install.

I didn't see anything via the link. Is it still available for download? Cheers

comemyfanatics:
> The first fully NDI shoot I did, I ran multiple wireless mics into a recorder/mixer and then out to a dedicated audio only camera. That went into OBS via NDI.
> 
> After much testing I now seperately run audio from wireless mics into a DAW (Ardour) via Dante AVIO hardware and then route the DAW output to the OBS machine, also using Dante, specifically Dante Via. OBS sees the audio as a sound card (WDM or ASIO). Dante Via and Dante virtual sound card handle the routing and syncing. The system works well, but is a bit costly if you include all the hardware... but it gives me incredible flexibility, professional audio interfaces - and it's fun to experiment. You don't need Dante hardware to make it work; Dante Via will route audio to/from other sound interfaces. Critically, Dante uses Precision Time Protocol to sync devices with very high precision and I suspect that's why its reliable.

How do you maintain sync between audio/Dante and multiple camera sources? I'm looking to do the same thing though with different equipment (Joeco Blackbox for audio, Birddog Flex 4Ks for videos). This may have an obvious answer though it's new to me.

aquarat:
Ah sorry, I only just realised this was directed at me. If I run Dante directly from the Dante AVIO ADCs to OBS using Dante Via (that software is insanely useful) I have set sync for the audio in the advanced audio properties of OBS to -130ms (!). All NDI sources were configured for Network timing and low latency. The sync remained stable for roughly 2 hours. This is for a simple two-mic setup.

For more complex setups, I run the Dante AVIO ADCs into Ardour DAW, then from Ardour into OBS running on a separate machine. This is all done with Dante Via. In this scenario I have to play with the latency settings in Ardour a tiny bit, but the setup is largely in sync and it doesn't drift over the course of roughly two hours.

On my next shoot I'm planning on using a Zoom L-8 for mixing instead of a purely software DAW. This is because the Dante AVIOs expect line-ish level input and my mics are a bit soft, so the Zoom L-8 will provide both preamps, mixing and recording facilities. I'll only use Dante to get the audio from the mixer out to OBS.

I've found that jitter becomes dangerous on Dante and NDI when you're running more than 4 HD streams on gigabit (at 100 mbit/sec per stream). I run 5 gigabit and 10 gigabit ethernet for the OBS machines.

Hope that helps.

PR #603, opened 2020-12-20
fix timecode translation

### LINKED ISSUE #539: OBS-NDI plug-in lagging on Windows
I'm using NDI Scan Converter to capture a full screen PowerPoint presentation running on a Windows 10 laptop.  On the Windows OBS computer, which is on the same Gigabit Ethernet network, I use the OBS-NDI plug-in to input the PowerPoint slides into OBS.  I've noticed that sometimes the NDI window in OBS lags behind by up to a second or two when I change PowerPoint slides.  When I play a video in PowerPoint, it comes through to OBS flawlessly.  When I run NDI Studio Monitor on the same computer as OBS, it keeps up with the PowerPoint slides and shows no lagging.  To illustrate the problem, I captured the following video [https://www.youtube.com/watch?v=H35Crg1Y6tc](https://www.youtube.com/watch?v=H35Crg1Y6tc).   In this video, recorded using OBS, I'm showing a window grab of NDI Studio Monitor in the top part and the OBS-NDI plug-in in the bottom part,.  You will see that sometimes the OBS-NDI plug-in lags behind, when switching to the next slide.

UPDATE:
The issue described above was fixed by changing 4 lines of code in obs-ndi-source.cpp where the NDI timestamp or timecode is translated to the OBS timestamp.  The NDI timestamp or timecode is in nanosecond intervals whereas the OBS timestamp is in 100 nanosecond increments.  In the existing code, the translation is done by multiplying by 100 (for example):
    obs_video_frame.timestamp = (uint64_t)(video_frame.timecode * 100);

The correct translation should be dividing by 100:
    obs_video_frame.timestamp = (uint64_t)(video_frame.timecode / 100);

This change should be made to all 4 lines of code in obs-ndi-source.cpp where the NDI timestamp or timecode is translated to the OBS timestamp.

mattatcha:
fixes #539

### DIFF
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -320,12 +320,12 @@ void* ndi_source_poll_audio_video(void* data)
 			switch (s->sync_mode) {
 				case PROP_SYNC_NDI_TIMESTAMP:
 					obs_audio_frame.timestamp =
-						(uint64_t)(audio_frame.timestamp * 100);
+						(uint64_t)(audio_frame.timestamp / 100);
 					break;
 
 				case PROP_SYNC_NDI_SOURCE_TIMECODE:
 					obs_audio_frame.timestamp =
-						(uint64_t)(audio_frame.timecode * 100);
+						(uint64_t)(audio_frame.timecode / 100);
 					break;
 			}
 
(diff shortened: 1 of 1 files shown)


mattatcha:
I'll need to debug this a little more since the audio from a video+audio source has a lot of crackling. I tried using the same timing translation for audio but that just resulted in no audio.

mattatcha:
Maybe something like this exists for the source. (Haven't looked into this at all yet.)

https://github.com/Palakis/obs-ndi/pull/564/files

Palakis (approved):

mattatcha:
@Palakis I think this should actually be reverted. I should have left this as a draft pr until I finished testing it :(

PR #609, opened 2021-01-16
fix tally light not working properly when multiview is enabled

### LINKED ISSUE #318: NDI Preview Tally sent if Multiview is opened.
Hello OBS-NDI-contributers,

NDI supports Tally-Lights for Preview and Program, so the camera can display if it is on air or about to go on air. OBS-NDI sends these Tally-Notifications, but it also sends a preview tally in multiview when the NDI Source is not in preview.

I have taken a few screenshots to visualize: https://imgur.com/a/wY3MyvX

Log (not that helpful, but anyways): https://obsproject.com/logs/5UST8pBwz6zkrs3K

Thanks in advance and have a nice day,
Kevin

dakriy:
fixes #318

### DIFF
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -418,7 +418,6 @@ void* ndi_source_poll_audio_video(void* data)
 
 bool ndi_source_in_preview(struct ndi_source* s)
 {
-
 	struct ndi_preview_check_context a;
 	a.found = false;
 	a.s = s;
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -21,6 +21,7 @@ along with this program; If not, see <https://www.gnu.org/licenses/>
 #endif
 
 #include <obs-module.h>
+#include <obs-frontend-api.h>
 #include <util/platform.h>
 #include <util/threading.h>
 #include <chrono>
@@ -70,6 +71,12 @@ struct ndi_source
 	NDIlib_tally_t tally;
 	bool alpha_filter_enabled;
 	os_performance_token_t* perf_token;
+};
+
+struct ndi_preview_check_context
+{
+	bool found;
+	struct ndi_source* s;
 };
 
 static obs_source_t* find_filter_by_id(obs_source_t* context, const char* id)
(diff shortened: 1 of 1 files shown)


norihiro:
I'm looking for the same feature for my plugin and I implemented it based on this PR.
I appreciate the idea. I'm glad if you can consider my comments.

https://github.com/Palakis/obs-ndi/blob/31837ce5feeebf28cc30efe4e0e351bd523b7adf/src/obs-ndi-source.cpp#L430-L438
Instead of `obs_scene_enum_items` and `obs_sceneitem_get_source`, how about using `obs_source_enum_active_sources`?
If my understanding is correct, the API `obs_source_enum_active_sources` will also enumerate nested sources such as a group and another scene in the preview scene.

https://github.com/Palakis/obs-ndi/blob/31837ce5feeebf28cc30efe4e0e351bd523b7adf/src/obs-ndi-source.cpp#L542
It looks too frequent to call `ndi_source_in_preview` for every tick.
How about register callback using `obs_frontend_add_event_callback` so that the enumerate API will be called only when the preview was updated.
I'd like to note that it would be author's choice because it requires an additional header file.

`decklink-output-ui` would be a good example. I put links to the code.
For my plugin `obs-command-source`, I put the code to register and remove callback at shown and hidden because my plugin will be instantiated a lot. In NDI-source, I think `ndi_source_create` and `ndi_source_destroy` would be the best place because the number of NDI sources would be not so many.
https://github.com/obsproject/obs-studio/blob/6942bb814dcf9361dc9c3dac395842444159be54/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp#L178-L179
https://github.com/obsproject/obs-studio/blob/6942bb814dcf9361dc9c3dac395842444159be54/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp#L125
https://github.com/obsproject/obs-studio/blob/6942bb814dcf9361dc9c3dac395842444159be54/UI/frontend-plugins/decklink-output-ui/decklink-ui-main.cpp#L209-L231

By the way, I recommend to [squash](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) the commit 31837ce into 2b18186 and force-push (`git push -f`) to github.

tt2468:
Closing this, as we won't be releasing another 4.x version of obs-ndi.

I have a different method of solving this, which is adding `obs_source_previewing()` to the OBS api, then using it instead of `obs_source_showing()`.

PR #618, opened 2021-01-29
Configurable tally

sampie777:
Added config options to enable/disable tally status on NDI source, as this status may influence source video display for other receivers (like NDI Studio Monitor).


src/obs-ndi-source.cpp
```diff
@@ -492,8 +493,9 @@ void ndi_source_update(void* data, obs_data_t* settings)
 			recv_desc.source_to_connect_to.p_ndi_name);
 
 		// Update tally status
-		s->tally.on_preview = obs_source_showing(s->source);
-		s->tally.on_program = obs_source_active(s->source);
+		Config* conf = Config::Current();
+		s->tally.on_preview = conf->TallyPreviewEnabled && obs_source_showing(s->source);
```
Palakis, line 497: Can you please have the tally configurable per-source? And have it take its default tally values from the default config.

Palakis (changes requested):
Thanks for this contribution! PR looks good to merge, though I only have one request before doing so:

src/obs-ndi-source.cpp
sampie777: @Palakis I can look into that. 
What do you refer to by "the default config"? As I think that the default values might just be `true` (as it's currently implemented in _master_), which require no lookup into some "default config".

Palakis (approved):

PR #665, opened 2021-06-12
generate a missing timestamp locally

sylphid8:


paulpv:
@sylphid8 I'll consider if this is still necessary after 4.11.0 is released

paulpv:
@sylphid8 Is this PR still necessary?

The current code for Sync `Network` vs `Source Timing` is:
https://github.com/obs-ndi/obs-ndi/blob/4daba126d0fc294a9f38213d9fb0a5a6b914c4ee/src/obs-ndi-source.cpp#L368-L380

I didn't write this code, so I am just saying as the current code maintainer that this option is obviously there for a reason.
Is the existence of this option enough to fix any issue you were seeing?
Is there anything that should or needs to change?

Trouffman:
This does not seems necessary according to the NDI Doc.
https://docs.ndi.video/docs/advanced-sdk/ndi-send#timecode-synthesis

`If you never specify a timecode at all [...], the current system clock time is used as the starting timecode [...], and synthetic values are generated.`

This is for "SEND" aka : anything that send an NDI stream will have a timecode (maybe that wasn't the case a long time ago).

This does not seems to fix any outstanding issues directly either. 

Also : no feedback from reporter.

PR #670, opened 2021-07-30
chinese simplified language update 2021/07/30

wwj402:
only language update


tt2468:
Closing this, since there will not be another 4.x version, and 5.0.0 will have all new strings.

Issue #720, opened 2022-02-17
Investigate/support NDI|HX 3


shleeable:
https://videoguys.com/blogs/news-and-sales/ndi-hx3-is-here-takes-ndi-hx-to-the-next-level

alessionossa:
I contacted people at NewTek / NDI and they said:

> NDI|XH3 is a format specification, you can create an HX3 stream with the actual SDK.
> 
> You simply implement an HX implementation as follows, 
> - GOP length of 1 or 2.
> - Bandwidth at a level where it is visually lossless with the HDMI/SDI output. (normally ~80Mbit/s)
> - Place The  video compressor in a low latency mode (e.g. I-frame only)
> 
> To claim NDI|HX3 support, NDI® must  test it and certify it In addition:
> - It must conform with other the latest usage implementation guide (e.g. for source naming, etc…)
> - It is visually lossless.
> - End to end latency with Studio Monitor is <100ms

high-solutions:
would be nice to add HX3 output option to support Gigabit networks with multiple streams. 
+1 for me.

ogmkp:
HX3 is just a tweak to prevent the artifacts from showing, we were already doing it on HX2 with more bitrate, they could have called it HX2.1, honestly in HEVC, 12 Mbits is enough for 1080p60.

PR #738, opened 2022-05-19
Placeholder for potential future HX|H264 and HX|HEVC - HX/HX2/HX3

paulpv:
ABSOLUTELY DO NOT MERGE!

This is just a non-compiling prototype stub for now **to get the conversation going**.
This most likely will also require a non-trivial update to NDI SDK v5.
It looks like some work for this may have already been done on some other forks or the `rewrite` branch.
https://github.com/Palakis/obs-ndi/tree/rewrite

Also related to #720 

@Palakis I see that @tt2468 started a rewrite from scratch on 2021/07/18 and you last pushed on 2021/11/19.
I perused the code a little bit and saw that the meat of NDI output code is still empty:
https://github.com/Palakis/obs-ndi/blob/rewrite/src/obs-ndi-output.cpp

I see that you have also been mega busy with the excellent osb-websocket v5 rewrite.
Zero pressure on these questions, I am just curious:
* Is the rewrite still active?
* Is the plan for it to support HX H.264 and HEVC encoding?
If so then this PR could probably be abandoned.

If not, my initial goal for this branch from the `master` is to (cough) "simply":
1. update the NDI lib to v5
2. add NDI 5 HX H.264 and HEVC output options

Other forks may have tried, and I am wondering if this effort would be well-served to attempt on `master`, or if anyone [still] attempting this would be better off waiting for the rewrite.

### DIFF
--- a/src/obs-ndi-output.cpp
+++ b/src/obs-ndi-output.cpp
@@ -256,6 +256,13 @@ void ndi_output_destroy(void* data)
 	bfree(o);
 }
 
+// Define this in a more appropriate place and make this a settable option in the UI
+typedef enum HxMode {
+	None,
+	H264,
+	HEVC
+} HxMode;
+
 void ndi_output_rawvideo(void* data, struct video_data* frame)
 {
 	auto o = (struct ndi_output*)data;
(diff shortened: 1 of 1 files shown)


paulpv:
@tt2468 Perhaps you already have more code fleshed out for https://github.com/Palakis/obs-ndi/blob/rewrite/src/obs-ndi-output.cpp#L29, but this something like this simple tweak may help to support HX|H264 and HX|HEVC output.
I think the plugin already handles receiving HX inputs.

(FYI, I'm the @NightVsKnight dude that was chatting with you earlier today on OBS Discord).

tt2468:
I'll have to look into it deeper, but IIRC the ability to send encoded data over HX requires the advanced SDK, which has questionable licensing for usage. Supporting HX output would require its own output_t type, as outputs cannot switch between asking for raw frames and encoded packets.

paulpv:
@tt2468 I was experimenting with creating a plugin from scratch using https://github.com/obsproject/obs-plugintemplate.
This is my progress so far:
https://github.com/NightVsKnight/obs-ndihx

I've so far only coded initializing NDI 5.5; it complies successfully but I have yet to actually load it in OBS.

Lots of plagiarism in the name of science, and I was also experimenting with forking the plugintemplate so that I can keep the template up to date with any of its tweaks, but I'm thinking now that it would be better if I had just downloaded the template and created a plugin with no prior commit history.

@Palakis 's, or hopefully an official first party plugin [although I can understand why OBS may not want to sanction a third party proprietary protocol], will always be the go to version, but I do want to experiment on my own with learning how to both use the NDI SDK [one day goal of writing both an Android and RPi Monitor, similar to https://orfast.com] and write an OBS plugin.

Worst case, if I both get NDI|HX working in my plugin, and NDI requires a license to use it, then my version/fork/whatever could be designed to allow someone to drop in their NDI license to get to use.

I am happy to keep my learning code pubic and credit your repo and let you/anyone plagiarize any of my code that may be useful.

alessionossa:
In a few places, such as https://github.com/obs-ndi/obs-ndi/pull/738#issuecomment-1206022386, https://github.com/obs-ndi/obs-ndi/issues/870#issuecomment-1479531488 and https://github.com/obs-ndi/obs-ndi/issues/745#issuecomment-1248661864 issues with license of NDI Advanced SDK are mentioned. However, looking at the licenses, I was not able to find the relevant differences. If this has been investigated, is it possible to have more details about what is problematic with the license?
Someone also mentions license fee (https://github.com/obs-ndi/obs-ndi/issues/870#issuecomment-1519196483) but as far as I know, there's no fee.

rse:
A side-note: although I also would like to see NDI|HX support directly from within OBS Studio, a decent workaround is to use the NDI Bridge application in its "Local" mode. It picks up the NDI streams of OBS Studio, re-offers them under its own namespace, and once requested it encodes from NDI to NDI|HX just fine. We use this approach just fine in our company filmstudio for sending a dozen streams over just a 1 Gbps ethernet link.

PR #862, opened 2023-03-07
Changing readme `Palakis` to `obs-ndi`

paulpv:
This is mostly just to confirm if I still have push access to `obs-ndi/obs-ndi` since it moved from `Palakis/obs-ndi`.


Palakis (approved):

paulpv:
I'll fix the badges later

PR #867, opened 2023-03-14
Add Latency buffering options

### LINKED ISSUE #851: [Bug]: Loss of frames from NDI-HX camera when "Low Latency (Experimental)" mode is selected
### Operating System Version

Windows 10

### OBS Version

Currently using 29.0.2 but was the same on earlier versions

### NDI Tools Version

5.5.3.0 (Any NDI5 version)

### Describe the bug

When the NDI-HX source is set to "Low Latency (Experimental)", OBS may drop every 3rd or 4th frame from the source causing jerky video. This happens intermittently, the source may run smoothly for a time before becoming jerky. After a time it may run smoothly again. Does not appear related to CPU or GPU load. Tested on isolated network with just NDI-HX camera, gigabit switch and OBS machine (no internet).

I enable this mode because in "Safe" latency mode, the latency appears to be variable which causes problems with audio sync.

### Steps to reproduce

This may be dependent on NDI source and/or OBS machine specs. I am using HuddlecamHD-NDI camera, Windows 10 laptop with i7, Nvidia 1650Ti graphics. May be different for full NDI and NDI-HX.

1. Set up camera source to be high bandwidth for testing (e.g. 1080p60 @ 16mbps CBR)
1. Create NDI source set to "Low latency" mode, no hardware acceleration.
2. Make a recording of a moving object on OBS.
3. Play back the recording on a player which can do single step frames 
4. Observe that some frames are repeated then there is a jump to the next frame.

### Expected behavior

All frames should be received as transmitted.

### Screenshots

_No response_

### Additional context

See https://github.com/Palakis/obs-ndi/discussions/848
This appears to be due to obs_source_set_async_unbuffered() being set to True when the "Low latency" mode is enabled.
If this is set to False, then the frame drop does not occur, at the cost of 2 or 3 frames more latency, but still better than "Safe" latency mode.

It would be good to add an option in the source select UI to allow this option to be enabled (for absolute minimum latency) or disabled (for low latency but safe). I have made a test version of this and could submit as PR. Or perhaps offer 3 options in the latency box, "Safe" "Low" and "Minimum (experimental)" where "Low" sets the NDI-SDK to low latency but keeps the OBS frame buffering enabled.

congoblue:
Add 3rd option to the source latency drop down to provide a middle option with OBS buffering still enabled, as frame drops prone to occur when this is disabled. Fixes #851.

NB: I have only tested this on Windows 10.


qwertychouskie:
This looks very beneficial, thanks so much for the improvement!  Hopefully it gets merged and released soon :)

congoblue:
> This looks very beneficial, thanks so much for the improvement! Hopefully it gets merged and released soon :)

It would be helpful if you would like to test this - please download test version on my fork of obs-ndi - you just need to copy the files into the data and obs-plugins folders.
https://github.com/congoblue/obs-ndi/actions/runs/4425025280

congoblue:
> This looks very beneficial, thanks so much for the improvement! Hopefully it gets merged and released soon :)

There is a test build of this change available at my fork of the plugin - please test and feedback how it works for you. You just need to download the appropriate "Artifacts" zip for your OS then copy the files into your OBS installation folder.
https://github.com/congoblue/obs-ndi/actions/runs/4425451666

paulpv:
I'll look at this soon.

congoblue:
> Hello, I downloaded the plugin version in your fork, but in the obs it doesn't show the third buffer option, what am I doing wrong? I couldn't find the Artifacts.zip you mentioned. Another detail, is it correct to enable Use Hardware Decoding, or leave it disabled? Thanks
You need to copy the contents of the "data" folder in the zip into your installed obs-studio "data" folder (mine is in c:/program files (x86)/obs-studio), and you need to copy the contents of the "obs-plugins" folder in the zip into your installed "obs-plugins" folder.

You should then get the 3 options in buffering.

I've been running this test version since March 15 and it works well for me. The Hardware Decoding option makes no difference for me but it may depend on the PC hardware you are using.

qwertychouskie:
Any updates on getting this merged?  Would be extremely useful :)

congoblue:
I have been running my test version for a few months now and there is still sometimes a buffering issue.
I can't work out how deep OBS's frame buffering is, but part of the code is written so that it always takes the frames oldest first and sometimes OBS can get way behind the source (like 15 or more frames). You can only fix this by restarting OBS, or by changing the NDI source then reselecting the original one, the buffering then goes back to normal. 

It would be better if the buffering was limited to a couple of frames because (at least in my application) the occasional glitch would be preferable to having the source lagging 15 frames behind.

paulpv (approved):

PR #885, opened 2023-05-09
Add ptz to ndi source

exeldro:
- removed OBS_PROPERTIES_DEFER_UPDATE to update the camera as soon as property is changed
- add check if ndi_name or bandwidth changed to recreate ndi_receiver
- hide source in audio mixer when no audio is selected
- add checkable ptz group to properties to control the pan tilt and zoom with sliders
image


paulpv:
@exeldro I can't explain how I didn't see this before.
I'll give it a review.
Are there any updates you have since May?
Does this require the NDI Advanced SDK?

exeldro:
@paulpv after this I added hotkeys, see:
https://github.com/obs-ndi/obs-ndi/compare/master...exeldro:obs-ndi:develop
I am not sure about NDI Advanced SDK, as far as I know I used the default SDK

src/obs-ndi-source.cpp
```diff
@@ -315,6 +337,17 @@ void ndi_source_getdefaults(obs_data_t *settings)
 	obs_data_set_default_bool(settings, PROP_AUDIO, true);
 }
 
+void ndi_source_ptz(struct ndi_source *s, obs_data_t *settings)
+{
+	if (!obs_data_get_bool(settings, PROP_PTZ))
+		return;
+	float pan = (float)obs_data_get_double(settings, PROP_PAN);
```
paulpv, line 344: Should there be any range checks here?
Will NDI ptz accept bad values and not bork?

src/obs-ndi-source.cpp
```diff
@@ -315,6 +337,17 @@ void ndi_source_getdefaults(obs_data_t *settings)
...
+	float tilt = (float)obs_data_get_double(settings, PROP_TILT);
+	ndiLib->recv_ptz_pan_tilt(s->ndi_receiver, pan, tilt);
+	float zoom = (float)obs_data_get_double(settings, PROP_ZOOM);
```
paulpv, line 347: Should there be any range checks here?
Will NDI ptz accept bad values and not bork?

src/obs-ndi-source.cpp
```diff
@@ -315,6 +337,17 @@ void ndi_source_getdefaults(obs_data_t *settings)
 	obs_data_set_default_bool(settings, PROP_AUDIO, true);
 }
 
+void ndi_source_ptz(struct ndi_source *s, obs_data_t *settings)
+{
+	if (!obs_data_get_bool(settings, PROP_PTZ))
```
paulpv, line 342: Should the code also test for `recv_ptz_is_supported`?

paulpv (commented):
Still looking through the other charges...

src/obs-ndi-source.cpp
exeldro: the min and max are set in the properties, but we can add an extra check here

paulpv:
> I am not sure about NDI Advanced SDK, as far as I know I used the default SDK

Sorry, acronymnesia: I was confusing Advanced KVM ability with Standard PTZ ability.

paulpv (approved):
Looks great.
I made a few changes in the `develop` branch `obs-ndi-source.cpp`, so I'll take care of porting your changes from master to develop after I merge your master based changes to master.

paulpv:
@exeldro Did you see any actual benefit to only reinitiating the ndi_receiver if the name, bandwidth, or hwacceleration changed?

The old code always reinitialized the ndi_receiver, and I am thinking of moving back to that unless you had an obviously good reason in your code.

exeldro:
@paulpv reinitiating the ndi_receiver is visible in the output, so if the user wants to do live changes viewers will see interruptions

paulpv:
> @paulpv reinitiating the ndi_receiver is visible in the output, so if the user wants to do live changes viewers will see interruptions

Shouldn't almost all of these options result in the preview resetting?
image

Do you think that only name, bandwidth, and hwaccel are the only properties that should reset the preview?

paulpv:
OH! I see how that could be a big issue with the PTZ preview!
Gotcha!
Maybe I'll tweak/invert the logic to not re-init the receiver if the change is just PTZ?

exeldro:
hwaccel does not need to reset because recv_send_metadata can be used to change that live and all options after that only change things local in obs so no need to reset the ndi reciever for that

Issue #904, opened 2023-08-09
[Bug]: Installing to system Library folder instead of home Library


feetstv:
### OBS Version

29.1.3

### NDI Tools Version

5.5.3

### Describe the bug

The `obs-ndi.plugin` bundle is installed to `/Library/Application Support/obs-studio/plugins` instead of to `~/Library/Application Support/obs-studio/plugins` (in one's home folder).

### Steps to reproduce

1. Install plugin using `obs-ndi-4.11.1-macos-universal.pkg`
2. Open OBS Studio, check the Tools menu
3. Plugin is not loaded

### Expected behavior

Plugin should be installed in correct directory and thus be loadable by OBS Studio

cernyjakub:
I can confirm this. I just installed the plugin on a new M2 macbook and after install, it was not loaded. 

After manually moving the folder `/Library/Application Support/obs-studio/plugins` into the `~/Library/Application Support/obs-studio`, is started to work.

ticoyaki:
@cernyjakub hey, ive been having the same issue and tried the thing you suggested

I moved the whole plugins folder to ~/Library/Application Support/obs-studio but it  still seems to be missing in my OBS link preferences
<img width="478" alt="Screenshot 2023-09-24 at 16 54 42" src="https://github.com/obs-ndi/obs-ndi/assets/145918581/96d34d33-6477-4a0c-b774-5d3d09dbc336">

Altho in my OBS, NDI source shows up already but it's not at all showing my nintendo switch screen from there. is there anything else that you did maybe?

Thanks!

paulpv:
@feetstv @cernyjakub @ticoyaki 

I will soon release a 4.12.0 version w/ updated NDI lib that might help/avoid this.
This version uses the updated obs-plugintemplate that installs the plugin using OBS' officially recommended practice.

You can try out 4.12.0 Release Candidate if you want.

The Pull Request https://github.com/obs-ndi/obs-ndi/pull/919 has been labeled with "Seeking Testers" and the build generates installer packages that you can download try out.

Steps:
1. Browse to https://github.com/obs-ndi/obs-ndi/pull/919
2. Click on "Show all checks"
3. Click on any of the "Details"
4. Click on "Summary"
5. Scroll down and see the "Artifacts"
6. Download the relevant Linux/MacOS/Windows artifact for your platform
7. Windows may complain that the download looks suspicious; you may safely ignore 
8. Unzip the download
9. Run the installer

[...]

PR #939, opened 2023-11-26
Fix the macOS plugin output path to install directly into the OBS Plugins directory.

0rangeFox:
Hey 👋, I believe it is currently safe to hardcode the path according to the [OBS Plugins guide](https://obsproject.com/kb/plugins-guide#install-or-remove-plugins). This approach will help us address the [#904](https://github.com/obs-ndi/obs-ndi/issues/904) issue, eliminating the need to manually execute commands to fix this seemingly simple "bug" where paths are not recognized correctly through the CMake Environment variables.

### DIFF
--- a/cmake/macos/defaults.cmake
+++ b/cmake/macos/defaults.cmake
@@ -33,7 +33,7 @@ set_property(CACHE CMAKE_OSX_DEPLOYMENT_TARGET PROPERTY STRINGS 13.0 12.0 11.0)
 # Use Applications directory as default install destination
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX
-      "$ENV{HOME}/Library/Application Support/obs-studio/plugins;~/Library/Application Support/obs-studio/plugins"
+      "~/Library/Application Support/obs-studio/plugins"
       CACHE STRING "Directory to install OBS after building" FORCE)
 endif()
 
--- a/cmake/macos/defaults.cmake
+++ b/cmake/macos/defaults.cmake
@@ -33,7 +33,7 @@ set_property(CACHE CMAKE_OSX_DEPLOYMENT_TARGET PROPERTY STRINGS 13.0 12.0 11.0)
 # Use Applications directory as default install destination
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX
-      "~/Library/Application Support/obs-studio/plugins"
+      "$ENV{HOME}/Library/Application Support/obs-studio/plugins;~/Library/Application Support/obs-studio/plugins"
       CACHE STRING "Directory to install OBS after building" FORCE)
 endif()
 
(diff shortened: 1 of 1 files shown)


paulpv:
@0rangeFox Thank you for the PR, but see my comment in https://github.com/obs-ndi/obs-ndi/issues/904#issuecomment-1826892262

In summary:
> I do not want to hard code "~/Library/Application Support/obs-studio/plugins" in the installation script just to get 1% of users working when the right thing for the installation script to do is to use "$ENV{HOME}/Library/Application Support/obs-studio/plugins" that it already is.

Sorry, but I am mostly tempted to close #904 By-Design/Won't-Fix.

paulpv:
Best path forward in my opinion is to add to a `Known Issues` section of the readme about using OBS Link and the known workarounds **to deal with OBS Link's** limitation/bug/issue...not obs-ndi's issue.
As a compromise, I will not close this Issue until that is added to the readme.

obs-ndi/obs-plugintemplate is designed to allow a user to install to either `~/Library/Application Support/obs-studio/plugins` or `/Library/Application Support/obs-studio/plugins`.
OBS Link seems to only be looking in `~/Library/Application Support/obs-studio/plugins`.
If a user installs to `/Library/Application Support/obs-studio/plugins` then OBS Link won't work.

This is not obs-ndi's problem/responsibility to fix by forcing 100% of its users to install to a hard code `~/Library/Application Support/obs-studio/plugins` path just to fix an OBS Link issue that only affects a tiny % of obs-ndi users.

0rangeFox:
After a good discussion and analysis on [#904](https://github.com/obs-ndi/obs-ndi/issues/904#issuecomment-1829519856), I've come to a conclusion and reverted the last commit, because the plugin itself + version of OBS indicated was no longer compatible with OBS Link, so this way it's safe to follow the official OBS guidelines.

paulpv:
> it's safe to follow the official OBS guidelines.

The official OBS guideline is:
https://github.com/obsproject/obs-plugintemplate/blob/2c39e8e447433be8c5246047675632432b3ba87a/cmake/macos/defaults.cmake#L36

All this PR does is change `$ENV{HOME}` to `~`.
`$ENV{HOME}` should already evaluate to `~`.
If it does not then it is a bug in obs-plugintemplate that affects all plugins based on obs-plugintemplate, so it should be fixed there.

Open a PR on obs-plugintemplate and convince them that this change is needed.

I try to regularly update this plugin with their latest code, so if they accept your change then we will pick it up then.

0rangeFox:
Well, before creating an Issue/PR unnecessarily on [obs-plugintemplate](https://github.com/obsproject/obs-plugintemplate), I opted to analyze it first, and I did some tests on my PC with that code `$ENV{HOME}/Library/Application Support/obs-studio/plugins` and got the following result, so I'd say that in CMake it's doing its job and it's on the right path. Now I just don't understand what's really interfering.

```bash
/Users/0rangefox/Applications/CLion.app/Contents/bin/cmake/mac/bin/cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_MAKE_PROGRAM=/Users/0rangefox/Applications/CLion.app/Contents/bin/ninja/mac/ninja -G Ninja -S /Users/0rangefox/Projects/CLion/Test -B /Users/0rangefox/Projects/CLion/Test/cmake-build-debug
-- /Users/0rangefox/Library/Application Support/obs-studio/plugins <- The result.
-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/0rangefox/Projects/CLion/Test/cmake-build-debug
```

So I'm wondering, is it just me who's having this problem? Or is it somewhere that's causing the output of the plugin build to be placed in a different place? I haven't checked all the code in the plugin template yet, but if there was something wrong with it too, it would be already fixed. 

Anyway, I'm going to close this PR because it doesn't make any sense.

paulpv:
@0rangeFox 
> Anyway, I'm going to close this PR because it doesn't make any sense.
Thanks for looking into it anyway!

PR #954, opened 2024-01-08
Bring back support for transparent backgrounds in NDI sources using latency mode low

nielsverhoeven:
In the release of obs-ndi version 4.12 a new latency mode was introduced (lowest (unbuffered)). With the introduction of this latency mode for both the low and lowest modes transparent backgrounds in NDI sources were no longer available. This pull request returns the ability to work with NDI source that contain a transparent background while using latency mode low.


paulpv:
@nielsverhoeven I will review this soon...

buildspec.json
```diff
@@ -49,7 +49,7 @@
         }
     },
     "name": "obs-ndi",
-    "version": "4.14.0",
+    "version": "4.13.1",
```
paulpv, line 52: Change this to 4.14.1, or preferably just leave alone.
This is on the develop branch and I will increment the version number to 4.15.0 anyway when actually released to master.

paulpv (approved):

PR #996, opened 2024-04-09
close audio thread

FanZR123:
https://github.com/obs-ndi/obs-ndi/issues/963

audio_output_open is called for dummy_audio_queue while no audio_output_close.
This makes audio_thread of NDI is still running after obs_shutdown.

It seems you forget call audio_output_close.


ogmkp:
Hi,
What about this PR ?

Thanks.

paulpv (approved):

PR #997, opened 2024-04-11
Update libndi-get.sh for Linux NDI library Install

Trouffman:
Updated the linux libndi install script helper

The original script worked on Debian (12.5) but not on Ubuntu 22.04 LTS.

This proposed version of the install script for the Libraries has been tested successfully on :
- Ubuntu 22.04.1 LTS 
- Debian 12 Bookworm

image

The script also work for NDI v6 (just need to change LIBNDI_INSTALLER_NAME for v6).

This should address the following issues : 
- https://github.com/obs-ndi/obs-ndi/issues/977
- https://github.com/obs-ndi/obs-ndi/issues/949

### DIFF
--- a/CI/libndi-get.sh
+++ b/CI/libndi-get.sh
@@ -4,32 +4,88 @@ set -e
 LIBNDI_INSTALLER_NAME="Install_NDI_SDK_v5_Linux"
 LIBNDI_INSTALLER="$LIBNDI_INSTALLER_NAME.tar.gz"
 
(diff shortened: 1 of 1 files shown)


Trouffman:
As this script "clean-up" behind it it breaks the libndi-package.sh script (is this still used ?)
I can look at an option for "nocleanup" passed to the script.

paulpv (approved):

PR #1000, opened 2024-04-14
Update README.md

Trouffman:
Update readme until full v6 compatiblity is released.
Include the fix for the libNDI_for_mac.pkg installation that fail since the release of NDI v6 (issue in NDI installer - already been reported to them)


paulpv:
The main reason I am hesitant to do this is because I want to move away from having to update the readme file for every little update.
I set up wiki pages several months ago to make this much easier to keep up to date without having to create pull requests to the repo.
https://github.com/obs-ndi/obs-ndi/wiki/1.-Installation

Trouffman:
I will go an update the wiki then. Keeping it there for now and will drop this once done.

Trouffman:
this is no longer needed

PR #1017, opened 2024-05-09
Fix bug 1003: windows installer missing from .zip 

BitRate27:
Move archive creation until after installer build. Also, output installer
to configuration folder.

### DIFF
--- a/.github/scripts/Package-Windows.ps1
+++ b/.github/scripts/Package-Windows.ps1
@@ -64,16 +64,6 @@ function Package {
 
     Remove-Item @RemoveArgs
 
-    Log-Group "Archiving ${ProductName}..."
-    $CompressArgs = @{
-        Path = (Get-ChildItem -Path "${ProjectRoot}/release/${Configuration}" -Exclude "${OutputName}*.*")
-        CompressionLevel = 'Optimal'
-        DestinationPath = "${ProjectRoot}/release/${OutputName}.zip"
-        Verbose = ($Env:CI -ne $null)
-    }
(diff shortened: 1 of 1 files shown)


paulpv:
> Move archive creation until after installer build.

What problem does this fix?
If this is broken then we would also want to fix https://github.com/obsproject/obs-plugintemplate/blob/master/.github/scripts/Package-Windows.ps1

I am mostly asking because I made the change to https://github.com/obsproject/obs-plugintemplate/blob/master/.github/scripts/Package-Windows.ps1 that got this to work.
https://github.com/obsproject/obs-plugintemplate/pull/101

I am curious what I missed.

paulpv:
> Also, output installer to configuration folder.

What problem does this fix?

BitRate27:
I think this was due to a misunderstanding on my part around the role of the .exe and .zip files.  After reading the installation instructions here https://github.com/obs-ndi/obs-ndi/wiki/1.-Installation I don't think this PR is needed and can be rejected. To clarify, I read the issue to be "the .exe file is missing from the .zip file". Now I see they are intended to be different install methods. You will also notice that trouffman closed the issue https://github.com/obs-ndi/obs-ndi/issues/1003 after he saw my PR. Sorry for the confusion.

paulpv:
Thanks. 
Also no need to output to configuration folder?

BitRate27:
No. I did that so the .exe would get included in the .zip. It is not needed or even expected in the .zip.

paulpv:
Closing per @BitRate27

PR #1025, opened 2024-05-20
Zombie references

Trouffman:
This aim to bring old references to older LibNDI, theses popped up for me, there might be some others.


paulpv:
What are `ZombieCalls` and `Zombie references`?

PR #1027, opened 2024-05-21
Audio only patch

Trouffman:
This should fix a bug that exist since.... 2017 (IIRC)

**BUT this is conflicting with some added new behavior when the source is enable but set to not keep the connection alive when not visible.**

Fix audio bug that audio-Only mode does not work when restarting OBS.

Main bug with workaround solution: https://github.com/obs-ndi/obs-ndi/issues/592

Should fix the issues :
https://github.com/obs-ndi/obs-ndi/issues/429
https://github.com/obs-ndi/obs-ndi/issues/578
https://github.com/obs-ndi/obs-ndi/issues/981
https://github.com/obs-ndi/obs-ndi/issues/906
https://github.com/obs-ndi/obs-ndi/issues/621
https://github.com/obs-ndi/obs-ndi/issues/467
https://github.com/obs-ndi/obs-ndi/issues/298

This was tested and confirmed fixing the bug on MAC OS on OBS 30.1.1.

### DIFF
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -539,9 +539,9 @@ void *ndi_source_thread(void *data)
 			     obs_source_ndi_receiver_name);
 
 			// Force a clean frame on receiver reset at all times.
-				obs_source_output_video(
-					obs_source,
-					config_most_recent.blank_frame);
+			obs_source_output_video(
+				obs_source,
+				config_most_recent.blank_frame);
 
 			if (ndi_frame_sync) {
 				ndiLib->framesync_destroy(ndi_frame_sync);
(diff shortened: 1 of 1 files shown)


paulpv:
How are we on this? Is this really the code/fix that we need, or were you still working or thinking on something else?

Trouffman:
It works except for the identified limitation outlined in the OP.

We should have this merged as it fixes quite a bunch of bugs, and the limitation is only for a new feature.

We should release both, this fix and the new feature BUT highlights the known issue below :

> Known bug : No Audio for the source after OBS restart if the NDI Source use "Disconnect when not visible" but source is set to "visible".

Trouffman:
This branch is out of date. And everything is now being fixed in #1043

PR #1032, opened 2024-05-31
CI: Remove quotes '' from receipt and version strings

chrisgrande:
macOS pkg builds after version 4.11.1 included quotes around the identifier and version numbers, effecting the packages receipts. This causes issues for software deployment system like Munki that are now unable to correctly compare installed versions.

pkg-receipt-version-quotes


paulpv:
I believe you that this is a problem.
My concern is that this project is [effectively] a fork of https://github.com/obsproject/obs-plugintemplate and any core generic obs-plugin build/packaging/installation problem should be fixed in https://github.com/obsproject/obs-plugintemplate first and once vetted and merged there could then be merged into this project.

That said, it wouldn't be the first time I have first added a fix here **and then** opened a PR on https://github.com/obsproject/obs-plugintemplate to have them get around to it whenever they can.
But frankly those few experiences that I have are from personal experience and blocking/slowing my development progress so I justified not taking the longer route.

Is this a **really** big and/or urgent problem?
If it is, then I can consider taking this.
If it only affects a few people then my preference is to wait for a proper fix in https://github.com/obsproject/obs-plugintemplate and then merge the result here.

chrisgrande:
No worries on the delay. It's really only an issue if you are using any type of deployment system to install the pkg and that system is using the receipts for installation checking/confirmation. 

I could open a PR on the parent project or defer to you doing it if you already have a rapport with them, what ever is smoother. It's a relatively minor change in the scheme of the project. Thanks so much.

Trouffman:
This should be suggested to the OBS-plugin-te plate for sure.

Here are result of my research:
At first I assumed that the single quote on identifier is there to all 'space' character to be used. BUT this is not allowed as a bundle I'd anyway ([source Apple Dev Help](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier) )

So for the identifier I do not see the reason for it being used.

For the version number this seems to be a similar story/behavior.

"version number
The version number (CFBundleShortVersionString) is a user-visible string that represents the bundle version with a required format of three period-separated integers (0-9).

Each integer provides information about the release in the format [Major].[Minor].[Patch]. For example, in 10.14.1, 10 is the major revision number, 14 is the minor revision number, and 1 is the maintenance release number. This string is used throughout the system to identify the bundle version and should match the version number you enter in App Store Connect."

So no need for quote as well it seems ^^

Trouffman:
This should be suggested to the OBS-plugin-te plate for sure.

Here are result of my research:
At first I assumed that the single quote on identifier is there to all 'space' character to be used. BUT this is not allowed as a bundle I'd anyway ([source Apple Dev Help](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleidentifier) )

So for the identifier I do not see the reason for it being used.

For the version number this seems to be a similar story/behavior.

"version number
The version number (CFBundleShortVersionString) is a user-visible string that represents the bundle version with a required format of three period-separated integers (0-9).

Each integer provides information about the release in the format [Major].[Minor].[Patch]. For example, in 10.14.1, 10 is the major revision number, 14 is the minor revision number, and 1 is the maintenance release number. This string is used throughout the system to identify the bundle version and should match the version number you enter in App Store Connect."

So no need for quote as well it seems ^^

paulpv:
@paulpv Note To Self to PR this to https://github.com/obsproject/obs-plugintemplate

PR #1043, opened 2024-07-21
Audio only master

paulpv:
This is a master based cherry pick of @Trouffman 's main commit in https://github.com/DistroAV/DistroAV/pull/1027 to give him credit for the fix.

This will be released as 4.14.1 in a few days.

### DIFF
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -497,9 +497,6 @@ void *ndi_source_thread(void *data)
 			case PROP_BW_AUDIO_ONLY:
 				recv_desc.bandwidth =
 					NDIlib_recv_bandwidth_audio_only;
-				obs_source_output_video(
-					obs_source,
-					config_most_recent.blank_frame);
 				break;
 			}
 			blog(LOG_INFO,
@@ -539,6 +536,10 @@ void *ndi_source_thread(void *data)
 			blog(LOG_INFO,
 			     "[obs-ndi] ndi_source_thread: '%s' Resetting NDI receiver...",
 			     obs_source_ndi_receiver_name);
+
+			// Force a clean frame on receiver reset at all times.
+			obs_source_output_video(obs_source,
+						config_most_recent.blank_frame);
 
 			if (ndi_frame_sync) {
 				ndiLib->framesync_destroy(ndi_frame_sync);
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -1025,7 +1025,7 @@ void ndi_source_update(void *data, obs_data_t *settings)
 							s->config.blank_frame);
 #if 1
 				blog(LOG_INFO,
-				     "[obs-ndi] ndi_source_update('%s'): Creating a clean frame on update"
+				     "[obs-ndi] ndi_source_update('%s'): Creating a clean frame on update",
 				     name);
 #endif
 			}
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -565,8 +565,7 @@ void *ndi_source_thread(void *data)
 			// Force a clean (blank) frame when settings change to Audio only
 			if (recv_desc.bandwidth ==
 			    NDIlib_recv_bandwidth_audio_only) {
-				obs_source_output_video(obs_source,
-							NULL);
+				obs_source_output_video(obs_source, NULL);
 #if 1
 				blog(LOG_INFO,
 				     "[obs-ndi] ndi_source_thread: '%s' Reset Frame for Audio Only",
@@ -919,8 +918,7 @@ void ndi_source_thread_stop(ndi_source_t *s)
 		s->running = false;
 		pthread_join(s->av_thread, NULL);
 		if (!s->config.remember_last_frame) {
-			obs_source_output_video(s->obs_source,
-						NULL);
+			obs_source_output_video(s->obs_source, NULL);
 		}
 	}
 }
@@ -1012,8 +1010,7 @@ void ndi_source_update(void *data, obs_data_t *settings)
 			if (config.bandwidth ==
 			    NDIlib_recv_bandwidth_audio_only) {
 				// Force a clean frame when source is updated
-				obs_source_output_video(obs_source,
-							NULL);
+				obs_source_output_video(obs_source, NULL);
 #if 1
 				blog(LOG_INFO,
 				     "[obs-ndi] ndi_source_update('%s'): Creating a clean frame on update for Audio Only mode",
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -1012,7 +1012,7 @@ void ndi_source_update(void *data, obs_data_t *settings)
 
 			if (config.bandwidth ==
 			    NDIlib_recv_bandwidth_audio_only) {
-				// Force a clean frame when source is updated 
+				// Force a clean frame when source is updated
 				obs_source_output_video(obs_source,
 							s->config.blank_frame);
 #if 1
--- a/src/obs-ndi-source.cpp
+++ b/src/obs-ndi-source.cpp
@@ -440,6 +440,14 @@ void *ndi_source_thread(void *data)
 
 	NDIlib_recv_create_v3_t *reset_recv_desc = &recv_desc;
 
+	if (!s->running) {
+		// Force a clean frame at first start
+		obs_source_output_video(obs_source, blank_video_frame());
+
+		blog(LOG_INFO, "[obs-ndi] +ndi_source_thread('%s'): Creating the first frame for NDI source",
+			obs_source_ndi_receiver_name);
+	}
+
 	while (s->running) {
 		//
 		// Main NDI receiver loop
@@ -537,15 +545,12 @@ void *ndi_source_thread(void *data)
 			     "[obs-ndi] ndi_source_thread: '%s' Resetting NDI receiver...",
 			     obs_source_ndi_receiver_name);
 
-			// Force a clean frame on receiver reset at all times.
-			obs_source_output_video(obs_source,
-						config_most_recent.blank_frame);
-
 			if (ndi_frame_sync) {
 				ndiLib->framesync_destroy(ndi_frame_sync);
 				ndi_frame_sync = nullptr;
 			}
 
+			// Reset ndi_receiver
 			if (ndi_receiver) {
 #if 1
 				blog(LOG_INFO,
@@ -574,6 +579,31 @@ void *ndi_source_thread(void *data)
 				break;
 			}
 
+			// Force a clean (blank) frame when settings change to Audio only
+			if (recv_desc.bandwidth == NDIlib_recv_bandwidth_audio_only) {
+				obs_source_output_video(obs_source,
+							blank_video_frame());
+#if 1
+				blog(LOG_INFO,
+					"[obs-ndi] ndi_source_thread: '%s' Reset Frame for Audio Only",
+					obs_source_ndi_receiver_name);
+#endif
+			}
+
+			/*
+			// TODO: Force a clean frame everytime we reset the receiver to ensure "clean" state. Unless "Remenber last frame" is enabled AND settings is not audio-only bandwidth.
+			if ( ADD-Remenber_last_frame_here || recv_desc.bandwidth == NDIlib_recv_bandwidth_audio_only) {
+				obs_source_output_video(obs_source,
+							blank_video_frame());
+#if 1
+				blog(LOG_INFO,
+					"[obs-ndi] ndi_source_thread: '%s' Reset Frame after settings change.",
+					obs_source_ndi_receiver_name);
+#endif
+			}
+			*/
+
+			// Apply Framesync Settings
 			if (config_most_recent.framesync_enabled) {
 				timestamp_audio = 0;
 				timestamp_video = 0;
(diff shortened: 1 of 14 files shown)


Trouffman:
There is a known issue in this PR that might collide with a V6 feature.

Let me test it on master before merging.

Trouffman:
Test Settings / steps:
- NDI Source set as "Audio Only"
- Restart OBS
- Start Recording

Results possibles:
- Success : Audio-Only Source is audible in recording when expected.
- Fail : Audio-Only Source does not behave as expected (ie.: missing on Recording)

Test results : 
| Behavior when hidden: | Keep source connected | Disconnect source when not visible |
| --------- | -------- | ------ |
| Keep last frame when disconnected : ticked | Success | Success |
| Keep last frame when disconnected : unticked | FAIL(1) | Success |

Failure status #1 : 
- Audio level is visible in UI
- Audio is NOT in recorded file / missing

image

Trouffman:
HOLD ON for merge!
Found a better way to address this issue and keep it compatible with other features.

Trouffman:
This is now compatible with the latest added features.

It should also fix some other issues where the "frame" of and NDI source is not visible (because size is 0x0 for some reasons).

This also should follow the concept that : 
- When you start : start with a clean state then apply settings

Tested on MacOS (M2) with OBS 30.1.2 + test Pattern Generator and a 1K tone

paulpv:
> This is now compatible with the latest added features.
> 
> It should also fix some other issues where the "frame" of and NDI source is not visible (because size is 0x0 for some reasons).
> 
> This also should follow the concept that :
> 
> * When you start : start with a clean state then apply settings
> 
> Tested on MacOS (M2) with OBS 30.1.2 + test Pattern Generator and a 1K tone

Related to the 0x0 frame, I have noticed in the latest OBS log output that it says:
```
info: [obs-ndi] +ndi_source_create('NDI™ Source'...)
error: bmalloc: Allocating 0 bytes is broken behavior, please fix your code! This will crash in future versions of OBS.
info: [obs-ndi] +ndi_source_update('NDI™ Source'...)
```

So we will need to address the 0x0 problem soon anyway! :)

paulpv:
@Trouffman Are we good/safe/ok to merge this?

Trouffman:
That error message is new it seems ! (The bmalloc 0 byte allocation)

This has been tested on Mac, not yet windows/Linux.

Can merge, will need testing anyway.

paulpv:
> Can merge, will need testing anyway.

This PR would be the place to test the changes on Linux and Windows.  
This PR will generate installers because of the `Seeking Testers` label.

Once merged to master there won't be any installers generated until an actual release is tagged.
I supposed we could tag a release and just not create a Release entry at https://github.com/DistroAV/DistroAV/releases and test that tag and once it is working then create a Release entry.

Trouffman:
Let do the test on this PR!
I can have windows tested soon.

Trouffman:
I quickly tested on Windows (11) but I need more people to test it on windows.

| Behavior when hidden: | Keep source connected | Disconnect source when not visible |
| --------- | -------- | ------ |
| Keep last frame when disconnected : ticked | tbc | tbc |
| Keep last frame when disconnected : unticked | tbc | fail |

src/obs-ndi-source.cpp
```diff
@@ -440,6 +440,16 @@ void *ndi_source_thread(void *data)
 
 	NDIlib_recv_create_v3_t *reset_recv_desc = &recv_desc;
 
+	if (!s->running) {
```
paulpv, line 443: I don't get this.
This line is:
```
if (!s->running) { ... }
```
The next immediate line after this if statement is:
```
while (s->running) { ... }
```

If the `if` statement is entered, then the while loop never will be entered.
If the `if` statement isn't entered, then the while loop will be entered. 

I know this `ndi_source_thread` runs inside of the while loop.

So, this tells me that this `if (!s->running) { ... }` statement is not doing anything.

😕

src/obs-ndi-source.cpp
```diff
@@ -440,6 +440,16 @@ void *ndi_source_thread(void *data)
...
+		// Force a clean frame at first start
+		obs_source_output_video(obs_source, blank_video_frame());
```
paulpv, line 445: This allocates a blank video frame and never releases it

src/obs-ndi-source.cpp
```diff
@@ -573,6 +581,19 @@ void *ndi_source_thread(void *data)
 				break;
 			}
 
+			// Force a clean (blank) frame when settings change to Audio only
+			if (recv_desc.bandwidth ==
+			    NDIlib_recv_bandwidth_audio_only) {
+				obs_source_output_video(obs_source,
+							blank_video_frame());
```
paulpv, line 588: This allocates a blank video frame and never releases it

paulpv:
Around https://github.com/DistroAV/DistroAV/blob/cb4007925b7cdefb3c7fc0fae729c1037f3a184c/src/obs-ndi-source.cpp#L202-L208
I looked into the `bmalloc: Allocating 0 bytes is broken behavior` log warning and it has been around for over 2 years.
We should probably fix this soon too! :)
```
static obs_source_frame *blank_video_frame()
{
	obs_source_frame *frame =
		obs_source_frame_create(VIDEO_FORMAT_NONE, 0, 0);
	frame->timestamp = os_gettime_ns();
	return frame;
}
```

This is causing the following OBS log message(s):
```
info: [obs-ndi] +ndi_source_create('NDI™ Source'...)
error: bmalloc: Allocating 0 bytes is broken behavior, please fix your code! This will crash in future versions of OBS.
info: [obs-ndi] +ndi_source_update('NDI™ Source'...)
```
* obs_source_frame_create: https://github.com/obsproject/obs-studio/blob/e215502b620ce9d97f4d0191adef24856682b9e9/libobs/obs.h#L2751-L2760
  * video_frame_init: https://github.com/obsproject/obs-studio/blob/8c71b0b586a0b0a21fb6de9113c57de168d28668/libobs/media-io/video-frame.c#L227
    * bmalloc: https://github.com/obsproject/obs-studio/blob/106c7aa61f3229be628a77ef840bce414580df82/libobs/util/bmem.c#L104-L107
      This warning has been there since 2022/07/16:  
      https://github.com/obsproject/obs-studio/blame/106c7aa61f3229be628a77ef840bce414580df82/libobs/util/bmem.c#L104-L107

paulpv:
I think we should hold off on this code change and not push it to `master`.
Any objections?

Maybe still keep it open to discuss?

paulpv:
@Trouffman I think we should hold off on this code change and not push it to master.
Any objections?

Maybe still keep it open to discuss?

Trouffman:
Agreed. This is not ready for prime time yet.

Trouffman:
So whenever I try to fix one behavior, this brings another side issue or another "combo" of settings that breaks

Could be about time to rework that part more deeply.

src/obs-ndi-source.cpp
Trouffman: agreed!

Trouffman:
@paulpv This would be the new audio-Only fix approach. 
Has been tested on : MACos (M2).

src/obs-ndi-source.cpp
```diff
@@ -1075,7 +1098,7 @@ void *ndi_source_create(obs_data_t *settings, obs_source_t *obs_source)
 		QString("OBS-NDI '%1'").arg(name).toUtf8();
 
 	// Allocate blank video frame
-	s->config.blank_frame = blank_video_frame();
+	// s->config.blank_frame = blank_video_frame();
```
paulpv, line 1101: Intentionally not creating any blank_frame?
Does this work?

src/obs-ndi-source.cpp
```diff
@@ -1006,6 +1017,18 @@ void ndi_source_update(void *data, obs_data_t *settings)
...
 				    obs_source_active(obs_source))) {
+
+			if (config.bandwidth ==
+			    NDIlib_recv_bandwidth_audio_only) {
+				// Force a clean frame when source is updated
+				obs_source_output_video(obs_source,
+							s->config.blank_frame);
+#if 1
+				blog(LOG_INFO,
+				     "[obs-ndi] ndi_source_update '%s'): Creating the first frame for NDI source on Update",
```
paulpv, line 1028: Remove the stray `):`

paulpv:
@Trouffman I will test this on Windows [and maybe Linux] once you reply to my comments.
That commented out `s->config.blank_frame = blank_video_frame();` makes me think you might have miss-committed/pushed this line.

src/obs-ndi-source.cpp
Trouffman: Correct, the creation of the blank frame moved to ndi_source_update().

src/obs-ndi-source.cpp
Trouffman: adjusted wording as well.

Trouffman:
Tested and confirmed it works & fix the bug on MACOS & Windows 11.

src/obs-ndi-source.cpp
paulpv: This is the only line where `.blank_frame = anything` is called.  
There are no other calls to `blank_video_frame()`.  
So, nothing else is ever creating a blank frame.  
Meaning all uses of `s->config.blank_frame` are just sending nullptr.

Maybe this is intentional.  
Maybe this is accidental.  
Maybe you stumbled on a solution accidentally?

If not calling `blank_video_frame()` is intentional and fixes this, then maybe the whole `static obs_source_frame *blank_video_frame()` method should be removed?

Trouffman: This is a happy little accident :D.

If we call "blank_video_frame()" = the bug exist. Setting config.blank_frame = nullptr (explicitly or not) fixes the problem.

I am looking though commit history to find the reason of that blank frame to see what it was used for. (fixing a bug maybe?)

Trouffman: Digging deeper :

This blank_frame was added to fix the Audio only that was "keeping the last frame/not cleaning the frame". 
https://github.com/DistroAV/DistroAV/commit/7d1cf135df9610ae244c32282fdb1c9b838a5aca

This was part of v4.3.0 : "Source bugfix: video not cleared when switching to Audio Only mode"
https://github.com/DistroAV/DistroAV/releases/tag/4.3.0

src/obs-ndi-source.cpp
```diff
@@ -573,6 +563,19 @@ void *ndi_source_thread(void *data)
 				break;
 			}
 
+			// Force a clean (blank) frame when settings change to Audio only
+			if (recv_desc.bandwidth ==
+			    NDIlib_recv_bandwidth_audio_only) {
+				obs_source_output_video(obs_source,
```
paulpv, line 569: Per https://docs.obsproject.com/reference-sources#c.obs_source_output_video
Just use:
```
obs_source_output_video(obs_source, nullptr);
```

src/obs-ndi-source.cpp
```diff
@@ -1006,6 +1009,18 @@ void ndi_source_update(void *data, obs_data_t *settings)
 	if (!config.ndi_source_name.isEmpty()) {
 		if (!s->running && (config.behavior == BEHAVIOR_KEEP ||
 				    obs_source_active(obs_source))) {
+
+			if (config.bandwidth ==
+			    NDIlib_recv_bandwidth_audio_only) {
+				// Force a clean frame when source is updated 
+				obs_source_output_video(obs_source,
```
paulpv, line 1016: Per https://docs.obsproject.com/reference-sources#c.obs_source_output_video
Just use:
```
obs_source_output_video(obs_source, nullptr);
```

src/obs-ndi-source.cpp
```diff
@@ -1074,8 +1089,8 @@ void *ndi_source_create(obs_data_t *settings, obs_source_t *obs_source)
 	s->config.ndi_receiver_name =
 		QString("OBS-NDI '%1'").arg(name).toUtf8();
 
-	// Allocate blank video frame
-	s->config.blank_frame = blank_video_frame();
+	// Initialize the blank frame config at Source's creation
```
paulpv, line 1092: Remove these two lines.

In fact, remove `ndi_source_config_t.blank_frame`

src/obs-ndi-source.cpp
Trouffman: Will update everywhere and get rid of the config.blank_frame then

paulpv:
@Trouffman Looks good! I will test this my tonight or tomorrow before approving the code.

paulpv:
Tests out fine on both Mac and Windows.
I tested combinations of:
* Bandwidth: Highest, Lowest, Audio Only
* Behavior: Keep source connected, Disconnect source when not visible
* Keep last frame when disconnected: On, Off
All behaved as I would expect it to.
More importantly:
1. I closed OBS
2. Reopened OBS
3. Immediately start recording/streaming
4. Wait ~10 seconds
5. Stop recording/streaming
6. Open recording and verify audio can be heard

I will merge to master.

Thanks @Trouffman !

Trouffman:
@paulpv Added clarity or the Bug fix :

BugFix : Audio is missing from audio-only NDI Source after OBS restart. 

Main bug with workaround solution: https://github.com/obs-ndi/obs-ndi/issues/592

Should fix the issues :
https://github.com/obs-ndi/obs-ndi/issues/429
https://github.com/obs-ndi/obs-ndi/issues/578
https://github.com/obs-ndi/obs-ndi/issues/981
https://github.com/obs-ndi/obs-ndi/issues/906
https://github.com/obs-ndi/obs-ndi/issues/621
https://github.com/obs-ndi/obs-ndi/issues/467
https://github.com/obs-ndi/obs-ndi/issues/298

PR #1044, opened 2024-07-22
6.0: 1 of 2: Rebrand to "DistroAV"

paulpv:
NDI6 will be added in PR 2 of 2.

This change introduces a problem if the user already has `obs-ndi` installed and then they install this code:
1. They will have both `obs-ndi` and `DistroAV` installed.  
    The user will have to know to uninstall `obs-ndi`.
    Maybe when `DistroAV` launches it should detect `obs-ndi` and prompt them to [migrate (see below) and] uninstall?
3. `DistroAV` will not see the user's `obs-ndi` settings.  
    The user will have to re-setup all of their settings. :/
    Maybe when `DistroAV` launches it should detect `obs-ndi` and migrate all `obs-ndi` user settings to `DistroAV`?

This may need to be solved in another PR. 😦 
A user would need to manually follow the uninstall instructions.
We would need to keep uninstall instructions specific to OBS-NDI so that people don't get confused and accidentally/unintentionally uninstall DistroAV.

The DistroAV plugin should not uninstall OBS-NDI during runtime.

Name any OBS Plugin that, during runtime:
* Detects another Plugin
* Uninstalls the other Plugin
-or even-
* Just Uninstalls itself

None do.

We might need to customize the Windows and Mac installer and, see if we can customize the .deb installer, to do this.

But, that still won't catch the large % of users that manually copy the files.


Trouffman:
Some thoughts on your PR :

1. We should not touch anything at runtime BUT check for it because that could create conflict.

Can this be an "uninstall script" that would NOT be at runtime but at "Installation" step. (Much easier to deal with (not in code!))

Basically adding a step : "to install distroAV, you must uninstall OBS-ndi first".

I can update the documentation as well.

2. I thought that we could access it and "migrate to distroAV" config.
I am not sure how this is doable but as we control both OBS-ndi and distroAV code, we should be able to do something like this.

- Read distroAV config - check for OBS-ndi config > import as distroAV config > wipe out the OBS-ndi config (or have it with a "obs-ndi-config-imported-to-distroav flag so this does not get re processed)


Or the new installation process migrate it is OBS-ndi is detected (might actually be easier ahaha).


Overall : Having this as a step in installation could be a good "middle ground" that prevent us from doing too much in the code. But deal with all this at the Install process.
Then we only check distroAV config in code. Period ^^

paulpv:
Migrating Settings may not be much of a concern at all!

Both config.cpp and all of the other settings (ndi-source, ndi-filter, main-output, preview-output, etc) load settings keyed with prefix `NDIPlugin`.
We don't need to rename that key, and both DistroAV and OBS-NDI can read and write to that same key with no problem.
image

Hindsight says that key **should** have been be a less generic namespace than `NDIPlugin`, something more like `org.distroav.obsndi` (that obsndi is fine, since the plugin is DistroAV's plugin for OBS and NDI)... but that is water under the bridge.
But, @Palakis was a psychic genius when he came up with that prefix because since it isn't `obs-ndi` we don't have to rebrand our configs/settings!

The only complication we may have is just how to best deal with loading 2 plugins.
They do both seem to be co-existing side by side... OK.

Interestingly, only one `NDI(tm) Source` shows up in the `Add Source` popup:
image

I'll look into which plugin that actually pops up (probably depends on which plugin loaded first or last), but maybe we should change that to say `DistroAV NDI(r) Source`?

Probably a similar story for the other properties (filter, etc).

But, as shown above, both plugins' global settings show up separately but load and save the same values.

NOICE!

Trouffman:
DistroAV NDI plugin is too long.
We could do the release that "nullify" the OBS-ndi part. (So when installing distroAV, it override the OBS-ndi and actively disable it)

Not sure how that would be detected but can be a simple message that display in the log file : OBS-ndi is now DistroAV, the OBS-ndi file will soon be removed from the releases in favor of distroAV file.

paulpv:
`DistroAV NDI® Source`  
is no longer than  
`macOS Screen Capture`  

I don't follow your comment...
> We could do the release that "nullify" the OBS-ndi part. (So when installing distroAV, it override the OBS-ndi and actively disable it)

How to "nullify" the OBS-NDI part?  
The only way I know to actively disable a plugin it is to just uninstall it, which the installer should do anyway when going from any version to the next.  
The problem is that not everyone will use an installer, so we have to have a non-installer solution, so we might as well make that solution a prominent and usable one.

DistroAV will simply detect OBS-NDI and if present tell the user something along the lines of:
1. OBS-NDI is now DistroAV.
2. DistroAV has detected a previous version of OBS-NDI.
3. DistroAV and OBS-NDI are separate plugins that use the same settings and should not be used at the same time.
4. DistroAV requires OBS-NDI to be uninstalled before it can load.
5. DistroAV will keep and use all of their OBS-NDI settings even after uninstalling OBS-NDI.
6. DistroAV will unload

> Not sure how that would be detected but can be a simple message that display in the log file : OBS-ndi is now DistroAV, the OBS-ndi file will soon be removed from the releases in favor of distroAV file.

What does ` the OBS-ndi file will soon be removed from the releases` mean?

paulpv:
Belay that!

I spoke a little prematurely after first pass of playing **a little bit** with the OBS UI when both plugins were loaded.

After playing with this a bit more, loading the DistroAV Settings and making and applying a change (Main Output on/off, name, etc), those changes do **NOT** show up in the OBS-NDI Settings, and vice versa.

I need to do a little more testing on this...

paulpv:
...confirmed things seem OK with settings.

The problem was simply that the `OutputSettings::showEvent(...)` was not calling `Config::Load()`.

Will be fixed in my next push.

Trouffman:
Some feedback on "avoiding conflict with OBS-NDI obsolete file" from my research :
- Installer versions : We should remove the old files there.
- Non-Installer version : Adding the step in the Wiki on how to manually remove (most of it are already there :D)

Anyway we need the check + error message if distroAV detect obs-ndi.

.github/ISSUE_TEMPLATE/BUG.yml
```diff
@@ -45,16 +45,16 @@ body:
   - type: input
     id: obsndi-version
```
Trouffman, line 46: Should this id be updated as well ?

buildspec.json
```diff
@@ -48,12 +48,11 @@
...
-    "displayName": "OBS-NDI",
-    "version": "4.14.0",
-    "website": "https://obsproject.com/forum/resources/obs-ndi-newtek-ndi%E2%84%A2-integration-into-obs-studio.528/",
-    "author": "Stéphane Lepin",
-    "email": "stephane.lepin@gmail.com",
+    "name": "DistroAV",
+    "version": "6.0.0",
+    "website": "https://distroav.org",
+    "author": "DistroAV",
+    "email": "distroav@distroav.org",
```
Trouffman, line 55: Wasn't it agreed to have contact@distroav.org ?

src/config.cpp
```diff
@@ -1,29 +1,28 @@
...
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU General Public License for more details.
 
-You should have received a copy of the GNU General Public License along
-with this program. If not, see <https://www.gnu.org/licenses/>
-*/
+	You should have received a copy of the GNU General Public License
+	along with this program; if not, see <https://www.gnu.org/licenses/>.
+******************************************************************************/
```
Trouffman, line 16: @paulpv Any reason for the fancy tab/blank-spaces at the line beginning?

(I personally do not mind the extra ***)

src/config.cpp
```diff
@@ -68,19 +68,22 @@ Config::Config()
 	  TallyPreviewEnabled(true)
 {
 	auto arguments = QCoreApplication::arguments();
-	if (arguments.contains("--obs-ndi-verbose")) {
+	if (arguments.contains("--DistroAV-verbose",
```
Trouffman, line 71: Suggestion : --distroav-verbose

src/ndi-source.cpp
```diff
@@ -870,8 +862,8 @@ void ndi_source_thread_process_video2(ndi_source_config_t *config,
 		break;
 
 	default:
-		blog(LOG_INFO,
-		     "[obs-ndi] warning: unsupported video pixel format: %d",
+		blog(LOG_WARNING,
```
Trouffman, line 865: @paulpv do you remember the argument for this upgrade to Warning?
Not sure if this was because we got some Issues report about it recently or not.

Trouffman (commented):
Could not find outlandish issues in the proposed changed.

.github/ISSUE_TEMPLATE/BUG.yml
paulpv: I see I missed a few renames.
I wasn't wearing my glasses when I was doing it initially and my regex was incorrectly `obs[-]ndi` instead of what it should have been as `obs(-?)ndi`.
Looks like I've got at least one more push to make!

paulpv:
> Some feedback on "avoiding conflict with OBS-NDI obsolete file" from my research :
> 
> * Installer versions : We should remove the old files there.
> * Non-Installer version : Adding the step in the Wiki on how to manually remove (most of it are already there :D)
> 
> Anyway we need the check + error message if distroAV detect obs-ndi.

For our wiki, we could also reference OBS' wiki:  
https://obsproject.com/kb/plugins-guide

buildspec.json
paulpv: Briefly discussed on Discord, but not formally agreed.

There are two different contexts for the two different uses of the email address.
1. buildspec.json : main standard "email address of the author of this plugin" type of "play store" contact info
2. copyright/legal headers

I considered all of the following (and already setup free email forwards for):
* code@distroav.org
* contact@distroav.org
* copyright@distroav.org
* dev@distroav.org
* distroav@distroav.org
* foss@distroav.org
* info@distroav.org
* legal@distroav.org
* license@distroav.org
* ...

I had already committed and pushed `distroav@distroav.org` when you mentioned in Discord to use `contact@distroav.org`

I personally liked the more warm/personable/approachable "my name is DistroAV and you can email me at `distroav@distroav.org`" over the more cold/formal/business of `contact@distroav.org`.

I can change it to `contact@distroav.org`, but I'd like a 3rd opinion on this fairly anchoring decision.  
@RayneYoruka , what do you think?

src/config.cpp
paulpv: I copied the tabbed look from OBS.
They are very scattered:  
https://github.com/search?q=repo%3Aobsproject%2Fobs-studio%20Copyright&type=code

I liked the margined look of the tabbed version best.  

Since I have to touch them all anyway, I just took the time to make them all consistent and look a little prettier to me.

I also refreshed the text from https://www.gnu.org/licenses/old-licenses/gpl-2.0.html#SEC4 . The only change was they provide a snail-mail address that I kept as a url.

src/ndi-source.cpp
paulpv: I changed the level to `LOG_WARNING` to match that the text says "warning".

.github/ISSUE_TEMPLATE/BUG.yml
paulpv: Resolving this; will be in my next push.

src/config.cpp
paulpv: Resolving this

paulpv:
CCing @Palakis on this one.

One last issue on this "Part 1 of 2" PR (basic rename/rebrand; no NDI6; NDI6 is "Part 2 of 2"):
https://github.com/DistroAV/DistroAV/blob/6.0_distroav_rebrand_and_ndi6/buildspec.json#L46-L50
```
    "platformConfig": {
        "macos": {
            "bundleId": "fr.palakis.obs-ndi"
        }
    },
```

This shows up on a Mac at:
`~/Library/Application Support/obs-studio/plugins/obs-ndi.plugin/Contents/Info.plist`
```
...
<key>CFBundleIdentifier</key>
<string>fr.palakis.obs-ndi</string>
...
```

To be obvious: The schema is `{company domain}.{app name}`

Two topics...

## Topic 1: What name to use?

I want to rename this to something like one of the following:
1. `org.distroav.distroav`
   This one is obvious, but **kindof** limits us to only ever releasing an app/plugin that works with only OBS and NDI.  
   Not that I think I will ever have the time to, but it would be nice if we did not immediately limit ourself and started off with a naming scheme that makes if possible to have apps/plugins that work with more than just OBS or NDI.
2. `org.distroav.distroav.obs-ndi`
   This makes it easier to fantasize about other hypothetical combos such as a `vmix-srt` (vMix already has NDI built in, which kills me why OBS thinks they need to stay away from officially supporting NDI).  
   The problem with this name is that is sticks with the taboo `OBS-NDI` that we are trying to get away from.
   But the fact is, how else would you call something that mashes up `foo` and `bar` other than `foo-bar`?
   Maybe...
4. `org.distroav.distroav.ndi4obs`
   Implies the display name "DistroAV: NDI for OBS", as if "DistroAV" is the app and "NDI for OBS" is a variant.
5. `org.distroav.distroavndi4obs`
   This name drops the `.` to imply a more mashed up name of "DistroAV NDI for OBS", as if that is just the whole name of the app.
   It is subtly different from #4.

It is just a package name that no one will ever really see, but to someone techy digging under the covers it could be important to disambiguate things.

And my fantasy of ever having more than one distroav variant is admittedly very ridiculous.

Still, a good flexible package name is an important thing to have.

I prefer # 4 or # 1.

I predict we will choose the simpler # 1.

So: What name to use?

## Topic 2: Does renaming this break anything?

I am not sure how changing will affect either:
1. Signing with the Apple Developer account.
   (Speaking of which, I should create a `DistroAV` Apple Developer account for next year)
   Is there any hard-code that the Apple Developer account is only set up to sign `fr.palakis.obs-ndi` and absolutely will not sign `org.distroav.distroav` and we have to set something else up?
3. Installing `org.distroav.distroav` when `fr.palakis.obs-ndi` is already installed.
   Will the Mac installer know these are the same app and how to handle the package rename?

Can Apple Developer signing or package installing handle package renames?

paulpv:
I pushed `org.distroav.distroav` just to get something on the board to discuss.
We can discuss this here or on Discord and change this any time up until we release 6.0.0.

src/ndi-source.cpp
paulpv: Resolving

paulpv:
I am going to go ahead and merge this PR to branch `6.0.0_actual`.
`6.0.0_actual` will be collection of all individually reviewed PRs that will be merged to `master` when everything is ready.

Issue #1064, opened 2024-08-04
NDI Source stop on Preview when “Disconnect when not visible” is enabled


Trouffman:
Not sure yet if this is a 4.14.1 or prior (when this function has been added).

Behavior (with the problem)
- open OBS in Studio Mode
- 2 Scenes - each with their own NDI Source (NOT duplicate) - both with "disconnect when not visible"
- Set Scene1 to Program, Scene2 to Preview
> NDI Source on Scene1 is okay
> NDI Source on Scene2 is stopped (not expected following my reading / understanding of the feature)

The problem can come from : 
- The wording is not clear on this feature, I considered this as a bug in my review of v6 for example, but it exist on previous 4.14.1 as well.)
- “Visible” >> My understanding is "seen in OBS/used" . As example : ”seen in preview” or "Used on a projector" means "is visible".
- This is NOT really a visible challenge when using OBS in non-studio mode (just the program, without preview)

I have a fix almost ready as well, finishing some testing before pushing the PR.

Trouffman:
This will be part of the next v^ release.

PR #1066, opened 2024-08-04
update disconnect feature behavior

Trouffman:
Fix the issue with the "disconnect stream when not visible" behavior

New behavior:
"Disconnect when source not visible" will stop the NDI Stream (saving bandwidth & ressources) when the Source is not visible on any Program/Preview/Display.

Previous behavior:
"Disconnect when source not visible" stop the feed as soon as it is not in Program / Stream.

Solve the following bug : 
https://github.com/DistroAV/DistroAV/issues/1064


paulpv (approved):

paulpv:
Thanks again @Trouffman !

paulpv:
@Trouffman does this honor `Keep last frame when disconnected` in both Preview and Program?

Trouffman:
> @Trouffman does this honor `Keep last frame when disconnected` in both Preview and Program?

It should. I will do more test to confirm.

PR #1071, opened 2024-08-08
source: clean up code/formatting and tweak logging

paulpv:
* ndi_source_thread now avoids multiple QByteArray comparisons
* ndi_source_thread  update obs_source_name if ndi_receiver_name changed


src/ndi-source.cpp
```diff
@@ -95,14 +104,41 @@ typedef struct {
...
+		  framesync_enabled(false),
+		  hw_accel_enabled(false),
+		  yuv_range(video_range_type::VIDEO_RANGE_PARTIAL),
+		  yuv_colorspace(video_colorspace::VIDEO_CS_709),
+		  latency(latency_),
+		  audio_enabled(true),
+		  ptz(),
+		  tally()
+	{
+	}
```
Trouffman, line 125: is this a typo ?

src/ndi-source.cpp
```diff
@@ -491,71 +535,82 @@ void *ndi_source_thread(void *data)
...
 			config_last_used.framesync_enabled =
 				config_most_recent.framesync_enabled;
 
-			reset_recv_desc = &recv_desc;
+			reset_ndi_receiver = true;
+
 			blog(LOG_INFO,
 			     "[DistroAV] ndi_source_thread: '%s' framesync changed to %s",
-			     obs_source_ndi_receiver_name,
+			     obs_source_name, //
```
Trouffman, line 551: is the "// " expected here ?

src/ndi-source.cpp
```diff
@@ -893,17 +959,26 @@ void ndi_source_thread_start(ndi_source_t *s)
...
+		auto obs_source = s->obs_source;
+		auto obs_source_name = obs_source_get_name(obs_source);
+		blog(LOG_INFO,
+		     "[DistroAV] ndi_source_thread_stop: '%s' Stopped A/V ndi_source_thread for NDI source '%s'",
+		     obs_source_name, s->config.ndi_source_name);
 		if (!s->config.remember_last_frame) {
-			obs_source_output_video(s->obs_source, NULL);
+#if 1
+			blog(LOG_INFO,
+			     "[DistroAV] ndi_source_thread_stop: '%s' Audio Only: Forcing a blank/clean video frame",
```
Trouffman, line 978: Actually this is just the wrong log message content.

"[DistroAV] ndi_source_thread_stop: '%s' Audio Only: Forcing a blank/clean video frame",

Actually this is just the wrong log message content.

"[DistroAV] ndi_source_thread_stop: '%s' : Forcing a blank/clean video frame",

(This here as nothing to do with Audio-only :D) 

Behavior : if the settings is to NOT keep the last frame. Wipe the texture when not visible anymore anywhere.

Trouffman (changes requested):
Some of the changes are clearly welcome for clarity! GG.

src/ndi-source.cpp
paulpv: No; that is intentionally an empty constructor.

src/ndi-source.cpp
```diff
@@ -95,14 +104,41 @@ typedef struct {
...
 	bool running;
 	pthread_t av_thread;
+
+	ndi_source_t()
+		: obs_source(nullptr),
+		  config(),
+		  running(false),
+		  av_thread()
+	{
+	}
```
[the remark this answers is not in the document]
paulpv: Intentional

src/ndi-source.cpp
paulpv: Yes; it forces the similar above and below lines of code all consistently line wrap nearly identically.

src/ndi-source.cpp
paulpv: Thanks for catching that.
I blame AI auto completion! :)

src/ndi-source.cpp
```diff
@@ -414,7 +445,7 @@ void *ndi_source_thread(void *data)
 	NDIlib_audio_frame_v3_t audio_frame3;
 	NDIlib_frame_type_e frame_received = NDIlib_frame_type_none;
 
-	NDIlib_recv_create_v3_t *reset_recv_desc = &recv_desc;
+	bool reset_ndi_receiver = false;
```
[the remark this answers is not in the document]
paulpv: git blame says I did it 2023/11/10:
https://github.com/DistroAV/DistroAV/commit/a07ad386b240d24adc8e57b7aef5de921f020d88#diff-6a31bbe65b845c97228452424d004830270b76230165c75ec3d1c67b2c6420e6R392

I remember doing it, but 9 months later, I saw that I was only using the `NDIlib_recv_create_v3_t *reset_recv_desc` struct being null or not null as simple boolean logic.
The struct reset_recv_desc served no other purpose than a boolean test, so I just replaced it with a simple boolean test.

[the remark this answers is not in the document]
paulpv: BUUUUUUT, a good catch that talking about this line made me realize this **SHOULD** be `bool reset_ndi_receiver = true;`.

The below code luckily logic handles setting this to true on the first pass, but it is safest/best to explicitly set to true.

I wil change.

PR #1072, opened 2024-08-08
build: Change plugin binary file from `DistroAV` to lowercase `distroav`

paulpv:


Trouffman:
Description of PR 
- Binary produced to be all lowercase - seems to be the norm of other files
- Add option to set a Displayed Plugin Name different than the Plugin Name used for binary and some other options.

Trouffman (approved):

Issue #664, opened 2021-06-12
NDI Source: "Allow hardware accelation" not working after start of OBS Studio


dj9pz:
I'm running OBS Studio 27.0.1 and obs-ndi 4.9.0 on Windows 10. I added an "NDI Source" to a scene and use "NDI HX Camera" on iPhone as a source.

"Allow hardware acceleration" is enabled, but seems to not be used after start of OBS Studio:

- "Allow hardware acceleration" is enabled
- Start OBS Studio, CPU usage is 14-17%
- Disable "Allow hardware acceleration" in the NDI Source and save
- CPU usage is unchanged at 14-17%
- Enable "Allow hardware acceleration" in the NDI Source and save
- CPU usage drops to 3-6% (and stays there)
- Restart OBS Studio
- CPU usage is back at 14-17%

Settings in NDI Source:
![image](https://user-images.githubusercontent.com/6215603/121772669-1774a800-cb77-11eb-9b83-27c6649d5f23.png)

Aerglonus:
_**off-topic**_  but is your HX Camera source showing properly? I've been trying to use NDI HX Camera with an iPhone 6 / iPad Pro 2019 NDI detects both devices but doesn't show anything not even a black screen

dj9pz:
@Aerglonus Yes, it is generally working, as long as there is proper network connectivity. In case of connection loss, I did encounter problems for it to reconnect. For my purposes (multi camera live streams) NDI HX Camera is not yet ready for "prime time" as it lacks stability and many features (remote control, white balance, fast reconnect) which DroidCam OBS offers.

tt2468:
This setting only tells NDI that it should use hardware acceleration if it thinks that it will help. It's unfortunately not any kind of guarantee.

JimmyPauwaert:
I confirm this issue. After OBS start the use hardware acceleration (for our NDI sources) are not honored. Just opening the source and saving fixes it. (Drops our cpu usage by 80% !)

ogmkp:
Not working here, no nvdec hardware acceleration with manipulations related here. OBS 29.1.3, Debian 12.
An NDI 5 SDK issue ?

tt2468:
> Not working here, no nvdec hardware acceleration with manipulations related here. OBS 29.1.3, Debian 12. An NDI 5 SDK issue ?

[...]

PR #1073, opened 2024-08-08
source: NDI5+ hardware acceleration

paulpv:
NDI5 changed the hardware acceleration request from `<ndi_hwaccel enabled="true"/>` to `<ndi_video_codec type="hardware"/>`.
We've never updated our code,
I'm updating our code. :)

Adding a long extensive comment in the code.

See also:
* https://docs.ndi.video/docs/sdk/performance-and-implementation#receiving-video
* https://forum.vizrt.com/index.php?threads/is-ndi_hwaccel-still-supported.251864/
* https://github.com/DistroAV/DistroAV/issues/664


paulpv:
@Trouffman As mentioned in Discord, this "feature" is a little silly and never worked in NDI5 and somehow we survived.

Options:
1. Use this PR's code as written when sends a working metadata when `Allow hardware acceleration` is checked and send a fake non-working metadata when it is unchecked.
2. Change this PR's code to only send the working documented metadata only when the `Allow hardware acceleration` is checked (ie: don't send a fake non-working metadata).
3. Not have any UI option and just always send a request to enable hardware acceleration.

I think an option is best.  
I could see some users wanting to legitimately reserve all of their GPU compute for gaming/whatever and not wanting any of their GPU used by NDI.

Trouffman:
Option 2.

paulpv:
I am tempted to turn this on by default, but I do not think that we should because [I think] the only way to directly turn this off for someone who does not want this enabled.

The only way I can think of to turn this off is to maybe reset the receiver, which is not intuitive how to do.  
We could make it so that **any** settings change stops and restarts the receiver.

One other way is to require the user to close and reopen OBS! 😦

Trouffman:
"Forcing" a receiver reset on a  settings change could be the way, this is already the case for some of them.

paulpv:
> "Forcing" a receiver reset on a settings change could be the way, this is already the case for some of them.

I have a solution for that in another PR, so I will go ahead and merge this relatively tiny PR.

PR #1074, opened 2024-08-09
6.0.0 "Visibility behavior" update (Text & logic)

paulpv:
This all started with https://github.com/DistroAV/DistroAV/pull/993

I love the feature!  
I even personally understand how it works and is supposed to be used.

But, it seems [to me] oddly difficult to expect other users to understand how to use it and other developers to understand how it is implemented (and how to maintain the code).  
It has even been the source of some tricky recent bugs. ☹️ 

The current UI is this:
image
```
* Behavior: (dropdown)
  1. Keep source connected
  2. Disconnect source when not visible
* [ ] Keep last frame when disconnected (checkbox)
```

That, at first, seems fine and mostly intuitive.

But when someone starts to use the feature things can get confusing.

* What does it mean to keep the source connected?
* What does it mean to disconnect a source?
* What is visibility? In Program or Preview or both?
* Is the checkbox related to the Behavior dropdown or not? They both say "disconnect".
* So is the checkbox being on [or off] saying that the last frame will be kept [or blanked] if there is **any** disconnect, or only if there is a disconnect due to visibility?
* So does the checkbox have any meaning if Behavior is `Keep source connected`?

I could go on with at least another half dozen or more questions.

I propose something like the following that, although currently ugly, does help, I think, clear up some confusion:
```
* Visibility Behavior: (dropdown)
  1. Not-Visible: Keep Active
  2. Not-Visible: Pause; Visible: Show Black Frame & Resume
  3. Not-Visible: Pause; Visible: Show Last Frame & Resume
```
image

This is what is implemented in this PR, but it mostly for discussion.

But, to implement [something like] this we would have to migrate the settings from the old/current values to the new values.
So that complicates the code a little bit.

What do y'all think?

CC @Trouffman @haakonnessjoen @RayneYoruka


Trouffman:
Still thinking about the best way to communicate that setting.
We could have a breaking change in V6. This is a relatively new feature. Not ideal but could solve some issues.

RayneYoruka:
It seems the cc got lost in my email.. woops! I like the proposed solution, it's clear and direct without being too cluttered or confusing

Trouffman:
I am growing more and more a fan of this drop-down menu. BUT with one change :

Change "Show Black Frame" by "Clean source. Content"

And "show last frame" to "Show last source content"

Why not using "frame"? 
- it is a technical term, not obvious for everyone at first (and this is the main issue this PR aim to solve)
- and "frame" is used in statistics for the encoder (output/recording), also in error messages for encoder issues like "frames missed to rendering lag" etc 

-> interestingly OBS does not use "frame" anywhere else, but refer to "Canva"/"boundary box", etc. which push me to not use frame as well. :D

src/ndi-source.cpp
```diff
@@ -547,6 +554,7 @@ void *ndi_source_thread(void *data)
 			}
 
 			// Deactivate the source output video texture when using Audio only
+			// ISSUE: Do we need to do both this and in `ndi_source_update`?
 			if (recv_desc.bandwidth ==
```
Trouffman, line 558: This is needed to "clean" the Source Content when switching to Audio Only i we want to keep the original "user experience" behavior as in :

- Open NDI Source settings
- Change bandwidth to audio only
> as soon as this is selected > it will "clear" the source content/frame
- On "save" > this should be calling the ndi_source_update ()

Ths goes along side with the UX we want for users : any changes made in the source settings shoud apply right away or only on save?
Default today is expected as : "apply as soon as changed". Commit to config on save. Revert on Cancel.

If we want to keep this > This should stay here.

src/ndi-source.cpp
```diff
@@ -953,6 +961,14 @@ void ndi_source_thread_stop(ndi_source_t *s)
 			"'%s' ndi_source_thread_stop: Stopped A/V ndi_source_thread for NDI source '%s'",
 			obs_source_name, s->config.ndi_source_name);
 
+		// ISSUE: I am not 100% convinced that we want to do this **here**.
+		// It depends on how we phrase/word the UI text.
+		// Do we really want to blank the video here **after** it stops,
+		// or should we really be blanking the video just **before** it restarts?
+		// Why can't we just blank the video in the receiver thread?
+		// Isn't that what the code did before the changes in this area?
+		// Why was that not sufficient? What were the problems with that?
```
Trouffman, line 970: Valid suggestion. IIRC on why this was moved here was because in previous commit this was keeping a black frame when the source was part of a scene in preview > but that was a bug as well.

Technically we should be okay to have this source texture "cleaned" when we restart the thread. - Will test and confirm

Trouffman:
Following up on the text itself, the VLC plugin is like this:
image

Which we could put like this:
image


On the behavior:
A simplified implementation is "Actively clean the frame on source update" - always EXCEPT "keep last frame and NOT audio-only"

Doing the update only in the "thread" > that require to add extra for managing the audio-only use-case.
Also tricky as it did not work well when changing the bandwidth while the source is hidden/inactive would not honor the setting either.

RayneYoruka:
> Which we could put like this: image

I like how this is, this is the most informative yet clean and easy to understand. We should push this

Trouffman:
In the last Commit on this PR before this message the key elements are :
- Clean Frame when starting (actually, on the source update) - works great so far (even when changing settings in an inactive source)
- Update of the text
- clean-up of the variable names to be more inline with the wording
- Management of the migration from 4.14.x option scheme to 6.0.x
- cleaning & Formating of the comments in the code when not necessary anymore.

paulpv:
> Following up on the text itself, the VLC plugin is like this:
image

Where did you find that?  
I looked in VLC after installing NDI Tools and could use the feature but could not find that UI setting.

Trouffman:
I really like the difference between pause and stop as well, hence why I kept them in the visible part.

The "stop" in the variable name is kept as it is closer to the actual "logic". So the inconsistency should be between the front facing and the internal logic for it only. (Aka just us, the devs/maintainers.

I don't mind changing it.

I don't like the large if/else if part for the configuration migration. > Maybe we should have a function at loading that is "migrate to new settings" where we only address the configuration changes (so the check happen only when loading the plugin (aka once), instead of everytime it is used. - not sure how that would work tbh.)

Trouffman (commented):
There is still abug that the NDI Srouce does not display when:
- OBS start
- Source is ONLY in a scene in Preview
- Source is not available (so not connected)

Still figuring out why this is a problem.

Trouffman:
@paulpv This is now feature complete and ready to be merged.

Tested and confirmed working on : 
- MacOS + OBS 30.1.1
- Windows + OBS 30.1.1

Trouffman:
Left to do is : update all other languages with the new wording.

Trouffman:
Some comment before the next commit. 

As @paulpv suggested we will present the following 3 options :
- Keepalive
- Stop & Reset
- Pause

"Pause" is technically a Stop with no Reset. But this to simplify UI. Behind the scene (no pun), this is actually stopping the receiver.

This also follow the same naming as the VLC plugin (which I assume they know better than me for wording).

Trouffman:
Confirming the following :

### New behavior options

- Keepalive
_Source stays connected even when not visible / not in-use_

- Stop & Reset
_Source disconnect when not displayed (preview/program/display) - Reset the last received frame when it becomes visible/used again_
NB: this can appear as fast "blink-once" for the source in slow environment

- Pause
_Source disconnect when not displayed (preview/program/display) - keep the last received frame when it becomes visible/used again_
NB: This can appear as a tiny "start-lag" for source in slow environment. If the source has not received anything since the last OBS start, this will be transparent/not visible.

### New Default Setting
The default behavior for any new NDI Source is : the "Pause" option.


### Migration from previous settings
Any NDI Source that have been configured / created in prior versions will migrate to the new behavior automatically as below:

- NDI Source had no specific "behavior" or custom behavior as "keep connected" settings ->  New setting = Keepalive
_This is consistent with the previous behavior and should be a transparent migration_
Note for future reviewer : the "keep last frame" has no effect with this use-case, so it is not checked.

- NDI source had a custom "disconnect" behavior (introduced in [4.14.0](https://github.com/DistroAV/DistroAV/releases/tag/4.14.0))
If "keep last frame" is ticked -> New setting = Pause
If "keep last frame" is not ticked -> New setting = Stop & Reset

src/ndi-source.cpp
```diff
@@ -1029,19 +1022,130 @@ void ndi_source_update(void *data, obs_data_t *settings)
 		(s->config.hw_accel_enabled != new_hw_accel_enabled);
 	s->config.hw_accel_enabled = new_hw_accel_enabled;
 
+	// Clean the frame content when settings change unless requested otherwise
```
Trouffman, line 1025: THis might be useful only for th eaudio only afterall.

Trouffman:
Seems recent changes on the main branch are impacting.

Known issue so far.

Issue 1 :

When an NDI source is connected to the NDI Pattern Generator with a static tone, there is no sound in audio / recording right after OBS launch.

image

This could be reproduced with other sources as well (inconsistently)
This happen when using "Pause" behavior


Issue 2 :
- Changing the Source name in the drop down (settings tab of a source) > This does not change the source right away, neither when clicking save. The source must be toggle on/off for this to apply.
This happen when using "Pause" behavior

Issue 3 : 
editing a scene, then clicking "cancel" will still reset the frame.
This happen when using "Reset"
NB : this does NOT happen if clicking the X icon to close. (seem part of the ""ndi_source_update"

12:02:43.179: [DistroAV] 'NDI 1' +ndi_source_update(…)
12:02:43.180: [DistroAV] 'NDI 1' ndi_source_update: Behavior Blank Frame: Deactivate source output video (Actively reset the frame content)
12:02:43.180: [DistroAV] ndi_source_update: 'NDI 1' behavior='stop_resume_blank'
12:02:43.180: [DistroAV] 'NDI 1' ndi_source_update: NDI Source 'DELLICIOUS (OBS)' selected.
12:02:43.180: [DistroAV] 'NDI 1' -ndi_source_update(…)


More being tested before this can be merged.

Trouffman:
Confirmed : these issues are from the main 6.0.0, not this one.

RayneYoruka:
> Seems recent changes on the main branch are impacting.
> 
> Known issue so far.
> 
> Issue 1 :
> 
> When an NDI source is connected to the NDI Pattern Generator with a static tone, there is no sound in audio / recording right after OBS launch.

It's really interesting I haven't caught this bug, specially with how much I've been messing with the audio via NDI lately (Like changing audio sources on the fly)

BitRate27:
It doesn't look like we are setting reset_ndi_receiver to true when the behavior changes. Is this on purpose?

Trouffman:
> It doesn't look like we are setting reset_ndi_receiver to true when the behavior changes. Is this on purpose?

Yes this is expected as this will be set when the source settings is confirmed, not when we pick from the dropdown.

The issue was because we were not checking the frame to remenber properly.

In this latest commit i had to get rid of all the "text-based" settings for the behavior and fallback to using int. It simplifies the codebase and it seems to work great, solving the challenges we had recently.

There should be some more clean-up to be done before release, but this should resolve most of the bugs we had identified for this release.

Trouffman:
The latest commit address issues when a change of ndi source name would not be detected properly and not reset the content on a change.
no
This has been tested locally (Mac) and seems to react properly to the changes in behavior options and source name.

Trouffman (approved):

Trouffman:
### New behavior options

- **Keepalive**
_Source stays connected even when not visible / not in-use_

- **Stop & Reset**
_Source disconnect when not displayed (preview/program/display) - Reset the last received frame when it becomes visible/used again_
NB: this can appear as fast "blink-once" for the source in slow environment

- **Pause**
_Source disconnect when not displayed (preview/program/display) - keep the last received frame when it becomes visible/used again_
NB: This can appear as a tiny "start-lag" for source in slow environment. If the source has not received anything since the last OBS start, this will be transparent/not visible.

### New default setting
The default behavior for any new NDI Source is : the "Pause" option.

### Migration from previous settings

**Any NDI Source that have been configured / created in prior v6.0.0 versions will migrate to the new behavior automatically to "KeepAlive"**

All "custom" behavior set in 4.14.x are reverted to KeepAlive - This is a breaking change.

patcreak (commented):
im confused

RayneYoruka:
> im confused

What are you confussed? Elaborate. 
**This pull was merged and closed months ago.**

What language is the part you've reviewed?

patcreak:
why i do with this, i try figure steam link

PR #1082, opened 2024-08-22
Use video_tick for source loop instead of a thread

BitRate27:
- Created a ndi_video_tick_context_t to store the local variables that were in ndi_source_thread. 
- Added ndi_video_tick_context_t tc to ndi_source_t.
- Set the video_tick callback in the obs_source_info to ndi_video_tick.
- Removed thread start and stop calls.
- Changed name of ndi_source_thread to ndi_video_tick.
- Removed local variables.
- Accessed old local variables indirectly through the s->tc.
- Changed break to running = false.
- Changed continue to return.
- Changed function names and debug comments from 'thread' to 'video tick'. 
- **Caution**: Had to increase number of samples when getting audio frame when using frame_sync from 1024 to 1600. This should be a computation based on number of seconds since the last video_tick and the sample rate of the audio which I couldn't find. 
- Noticed no audio when !frame_sync, but this was not working before this change either.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -115,10 +115,41 @@ typedef struct ndi_source_config_t {
 	NDIlib_tally_t tally;
 } ndi_source_config_t;
 
+typedef struct ndi_video_tick_context_t {
+	ptz_t ptz;
+	NDIlib_tally_t tally;
+
+	obs_source_audio obs_audio_frame;
+	obs_source_frame obs_video_frame;
+
+	NDIlib_recv_create_v3_t recv_desc;;
+
+	NDIlib_recv_instance_t ndi_receiver;
+	NDIlib_video_frame_v2_t video_frame2;
+
+	NDIlib_metadata_frame_t metadata_frame;
+	NDIlib_framesync_instance_t ndi_frame_sync;
+	NDIlib_audio_frame_v2_t audio_frame2;
+	int64_t timestamp_audio;
+	int64_t timestamp_video;
+
+	NDIlib_audio_frame_v3_t audio_frame3;
+	NDIlib_frame_type_e frame_received;
+	const char *obs_source_name;
+	ndi_video_tick_context_t()
+	{
+		memset(this, 0, sizeof(*this));
+		obs_audio_frame = {};
+		obs_video_frame = {};
+		recv_desc.allow_video_fields = true;
+		frame_received = NDIlib_frame_type_none;
+	}
+} ndi_video_tick_context_t;
+
 typedef struct ndi_source_t {
 	obs_source_t *obs_source;
 	ndi_source_config_t config;
-
+	ndi_video_tick_context_t tc;
 	bool running;
 	pthread_t av_thread;
 } ndi_source_t;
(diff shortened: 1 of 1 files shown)


BitRate27:
Also moved `auto config = Config::current()` from start of old thread to just before it is used for checking tally enabled.

src/ndi-source.cpp
```diff
@@ -115,10 +115,42 @@ typedef struct ndi_source_config_t {
...
 
+typedef struct ndi_video_tick_context_t {
+	ptz_t ptz;
+	NDIlib_tally_t tally;
+
+	obs_source_audio obs_audio_frame;
+	obs_source_frame obs_video_frame;
+
+	NDIlib_recv_create_v3_t recv_desc;
+	;
```
Trouffman, line 126: typo here ?

paulpv:
@BitRate27 this is a beautiful change, but I am leaving on holiday in a few hours and need some time to think about these changes.
I probably will not be able to focus on this until I get back ~9/3.

paulpv:
@BitRate27 Can you give a summary here of any performance metric improvements of using this technique?

paulpv:
@BitRate27 It would be awesome to be able to A/B test this live in the same build; have both the previous thread way and this new video_tick way.

FYI, OBS link to `obs_source_info.video_tick`:
https://docs.obsproject.com/reference-sources#c.obs_source_info.video_tick

Trouffman:
I am struggling to understand that part of the video_tick purpose, any documentation that would address the overall advantages?

Does this work when it is audio only ?

BitRate27:
The original code used a separate thread to capture the frames and send them on to the source API for further processing by OBS. This thread was a loop constantly capturing frames and passing them along. The time between capturing frames depended on how long it took to go through the loop. This time could vary depending on processor speed, how much data to process, and various settings. The theory is that if the loop takes less that 33ms (30 fps), we are doing extra work in the loop as the frame didn't change. 

To be the most efficient possible, we should only capture and send a frame every 33ms. Fortunately OBS has a callback that is called at this interval called `video_tick`. In summary, I took the loop out of the thread function and ask OBS to call the function every 33ms.

BitRate27:
> @BitRate27 Can you give a summary here of any performance metric improvements of using this technique?

I have done some tests on a debug version on a very old Windows box and do not see a decrease in CPU utilization. I suspect this is because the video_tick execution time is up close to the fps. So more tests to come.

Trouffman:
> The original code used a separate thread to capture the frames and send them on to the source API for further processing by OBS. This thread was a loop constantly capturing frames and passing them along. The time between capturing frames depended on how long it took to go through the loop. This time could vary depending on processor speed, how much data to process, and various settings. The theory is that if the loop takes less that 33ms (30 fps), we are doing extra work in the loop as the frame didn't change. 
> 
> To be the most efficient possible, we should only capture and send a frame every 33ms. Fortunately OBS has a callback that is called at this interval called `video_tick`. In summary, I took the loop out of the thread function and ask OBS to call the function every 33ms. 
> 

Thanks! That make so much more sense. One of my concerns was to get as much of the of /else and update of settings out of the main loop to limit performance impact. Seems this is taking the processing out of the loop, for the same purpose :D !

BitRate27:
@paulpv @Trouffman I am not seeing any performance increase with this change. I was also hoping this would fix the sync offset changing through a broadcast, but it didn't fix it either. More serious is that the audio is not working when not in frame_sync mode. I suggest we shelve/deny this PR. Will investigate other ways to increase performance and keeping a constant sync offset.

paulpv:
@BitRate27 Good choice. Sometimes better code doesn't equal improvement. Thanks for trying, and keep trying.

Issue #687, opened 2021-09-27
NDI Tally Light Issues...


Jaxel:
I'm using a BirdDog PF120 over NDI. It has a Tally light on it that turns on when the camera is triggered. The problem I am having is with the state of the tally light...

When the camera goes live, the tally light rapidly blinks between red and green. It shouldn't do this, it should be a solid color; ideally solid red.

The interesting thing is, if I put OBS in studio mode, it's the same, red and green in "program". However, if the camera is not in the live program feed, but is queued up in the preview feed, that's when the tally light is solid green.

How do I fix this? It should never rapidly blink between red and green.

It should be:
* Preview Queue = solid green
* Program Feed = solid red

![giphy](https://user-images.githubusercontent.com/3640402/134843607-6d22dd99-d1ad-438e-a087-21873766b320.gif)

tt2468:
We don't do anything special with tally, I don't see why it would do this as a result of something we're doing, we handle tally in the most by-the-book way possible.

paulpv:
Try this in the latest `develop` branch `4.11.0-RC` tagged version that is due to officially release once approved: https://github.com/Palakis/obs-ndi/pull/849

BitRate27:
I have duplicated this issue here: 
![TallyDouble](https://github.com/user-attachments/assets/31492abc-5d02-453a-9afc-2dda97d37c1a)

As you can see in Studio Monitor, when a NDI source in on program and not on preview, we turn both tallies on. 
I have a fix for this.

Trouffman:
Heads up on this initial bug report & the fix :

There are 2 elements here :

Behavior 1 :
- The PF120 blink red & green : This could be okay if the camera is in preview AND Program (I am confirming with Birddog about this)

Behavior 2 :
- DistroAV still send the preview + program as true when a source was in preview then moved tot Program, EVEN when the source is not in preview anymore.

> this mean the "tally_on-preview" is not cleared when the scene is moved to program only.
This is also the fix from @BitRate27 try to address

Trouffman:
Update on this : 
* Seems this _could_ be fine from a pure NDI perspective to send BOTH on_preview & on_program metadata if the sourc is actually in both.

This means that the camera is free to honor either:
* both signal (like on the NDI Studio it appears as 2 tally)
* the highest signal (on_program)
* the latest received (not a good practice in this case here)

Seems behavior 2 is the only part we need to fix afterall (or I am missing something)

Trouffman:
[...]

PR #1083, opened 2024-08-29
Fix tally sent to ndi devices to properly reflect status in OBS

BitRate27:
This fix addresses the problem identified in https://github.com/DistroAV/DistroAV/issues/687# by making the following changes:
1 - Add a signal callback on all ndi sources to react to scene changes and set tallies appropriately.
2 - Implement a function to determine if a source is shown in preview scene.
3 - Only set program tally in activated/deactivated callbacks.
4 - Only set preview tally in shown/hidden callbacks.
5 - Only check Tally enabled when deciding whether or not to send the tally info to the NDI device.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -973,6 +973,25 @@ int safe_strcmp(const char *str1, const char *str2)
 	return strcmp(str1, str2);
 }
 
+bool source_showing_in_scene(obs_scene_t *scene, obs_source_t *source)
+{
+    std::map<obs_source_t*, bool> source_map;
+    source_map[source] = false;
+
+    obs_scene_enum_items(scene, [](obs_scene_t*, obs_sceneitem_t* item, void* param) -> bool {
+        auto source_map = static_cast<std::map<obs_source_t*, bool>*>(param);
+        obs_source_t* source = obs_sceneitem_get_source(item);
+        auto it = source_map->find(source);
+        if (it != source_map->end()) {
+            it->second = true;
+            return false;
+        }
+        return true;
+    }, &source_map);
+
+    return source_map[source];
+};
+
 void ndi_source_update(void *data, obs_data_t *settings)
 {
 	auto s = (ndi_source_t *)data;
@@ -1070,9 +1089,10 @@ void ndi_source_update(void *data, obs_data_t *settings)
 
 	// Update tally status
 	auto config = Config::Current();
-	s->config.tally.on_preview = config->TallyPreviewEnabled &&
-				     obs_source_showing(obs_source);
-	s->config.tally.on_program = config->TallyProgramEnabled &&
+	auto preview_scene = obs_frontend_get_current_preview_scene();
+	s->config.tally.on_preview =
+		source_showing_in_scene(obs_scene_from_source(preview_scene),obs_source);
+	s->config.tally.on_program =
 				     obs_source_active(obs_source);
 
 	if (strlen(s->config.ndi_source_name) == 0) {
@@ -1125,7 +1145,7 @@ void ndi_source_shown(void *data)
 	auto s = (ndi_source_t *)data;
 	auto obs_source_name = obs_source_get_name(s->obs_source);
 	obs_log(LOG_INFO, "'%s' ndi_source_shown(…)", obs_source_name);
-	s->config.tally.on_preview = (Config::Current())->TallyPreviewEnabled;
+	s->config.tally.on_preview = true;
 	if (!s->running) {
 		obs_log(LOG_INFO,
 			"'%s' ndi_source_shown: Requesting Source Thread Start.",
@@ -1156,7 +1176,8 @@ void ndi_source_activated(void *data)
 	auto s = (ndi_source_t *)data;
 	auto obs_source_name = obs_source_get_name(s->obs_source);
 	obs_log(LOG_INFO, "'%s' ndi_source_activated(…)", obs_source_name);
-	s->config.tally.on_program = (Config::Current())->TallyProgramEnabled;
+	s->config.tally.on_program = true;
+	s->config.tally.on_preview = false;
 	if (!s->running) {
 		obs_log(LOG_INFO,
 			"'%s' ndi_source_activated: Requesting Source Thread Start.",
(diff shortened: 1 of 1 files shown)


Trouffman:
I was wondering if only updating in the activated/shown trigger would have work.

While at it we should add the option to send tally update (or not) for each source.

BitRate27:
After analyzing the further with @Trouffman here[ ](https://github.com/DistroAV/DistroAV/issues/1086), a new approach of using scene events instead of source events proves more robust and complete in setting tally. Here is the reasoning behind the new approach and and the resulting logic.
In OBS, the user can toggle the visibility of a source in a scene. In order to handle tally correctly, we need to tie preview and program status to the scene item and use the scene APIs instead of activate/deactivate/show/hide source APIs. The source APIs are good for sources to determine how it is being used, however since a source can be in multiple scenes and even be in a projector, the source APIs are not useful for determining preview or program visibility.

on_preview = source visible in current preview scene
on_program = source visible in current program scene

We should take action on setting the tally by registering a callback for a ndi source using obs_frontend_add_event_callback passing the callback data as ndi_source_t. This callback will be called when the current preview scene or program scene changes and will supply our ndi_source_t structure.

If the preview scene changes:

- get the current preview scene (obs_frontend_get_current_preview_scene)
- go through all scene items (obs_scene_enum_items)
- if the source is in the scene (obs_sceneitem_get_source) and it is visible (obs_sceneitem_visible) set the on_preview tally of the source to true, otherwise set it to false.

If the program scene changes:

- get the current program scene (obs_frontend_get_current_scene)
- go through all scene items (obs_scene_enum_items)
- if the source is in the scene (obs_sceneitem_get_source) and it is visible (obs_sceneitem_visible) set the on_program tally of the source to true, otherwise set it to false.

There is a signal on scene items that communicates item visibility change: https://docs.obsproject.com/reference-scenes#scene-signals (item_visible (ptr scene, ptr item, bool visible)). We should also respond to this signal and set the tally.

If scene item is a ndi source, the scene is in preview, and the scene item is visible then set on_preview to true, otherwise set it to false.
If scene item is a ndi source, the scene is in program, and the scene item is visible then set on_program to true, otherwise set it to false.

This logic works when the source exists multiple times in a scene. If any scene item for a source is visible, then the appropriate tally is on.

paulpv:
@BitRate27 ready to look into this more now that 6.0.0 has been released?

Trouffman:
@paulpv This issue technically still exists but we where looking to find a better way to detect if "a source is in use _at all_ anywhere as Preview / Program".

It was too much of a change for 6.0.0 and too many possible things would be breaking (or over complexify the issue debugging) if merged to 6.0.0.

I would prefer and official / recommended / supported / uniformed way to have the "source's state" to be computed (from official OBS's hook) that will help :
* long term support
* easier debugging/change management
* easier contribution by current & future maintainer / PR

We have the following options ([Source OBS Docs](https://docs.obsproject.com/reference-sources#c.obs_source_info.activate))
```
activate (ptr source)
    Called when the source has been activated in the main view (visible on stream/recording).

deactivate (ptr source)
    Called when the source has been deactivated from the main view (no longer visible on stream/recording).

show (ptr source)
    Called when the source is visible on any display and/or on the main view.

hide (ptr source)
    Called when the source is no longer visible on any display and/or on the main view.
```

BitRate27:
This PR works around an API that is missing from OBS Studio. It uses the scene change signal and traverses scenes to find its own source in preview and/program scenes.

I would propose 4 new callbacks added to OBS Studio API, like the show/hide and activate/deactivate, called on_preview/off_preview and on_program/off_program. This way obs keeps track and we just react. I feel it is an important part of a video broadcasting system API. 

The show/hide and activate/deactivate have their own purpose and changing how those work would be breaking.

If we put this fix in as is, we can always implement the new APIs when they show up, and remove the scene signal and traversal code.

Trouffman:
Thank you @BitRate27 & @bgillock for this PR.

After revising this PR, there are some interesting challenges to provide a more "precise" control over tally. 

Namely identify when a source is:
- enabled on PGM / STREAM
- enabled on PREVIEW
- enabled on Multiview
- enabled on any display

As of today there is not distinction for: Preview / multiview / any display. 

This is a complex issue as it will depends on how people are using their NDI Source.

ie.: should a source where and NDI Filter (output) is used also consider updating the Tally ? - depends

ie.: should a source not visible (hidden) BUT on a scene in Program be considered : OFF / On Preview or... depends ? (example when people use hide/show in OBS non studio-mode)


The initial bug this is supposed to address has had more information that seems to reduce the importance of the change as well.

From a pure code-maintenance perspective. I do not think the project is there yet to implement & maintain a whole new way to detect status of a source.

This might be better geared toward OBS Studio development suggestion. Why is this not already part of it? We might be missing something or not deemed enough interest.

There are similar suggestion for similar needs :
- https://ideas.obsproject.com/posts/1271/add-obs_source_inpreview-to-sdk


What to do next? 
- Define the goal of the Tally and its usage as would fit the most user/use-cases
- Address possible bugs not clearing _on-preview_ on change/ close of the source connection
- Put this PR on-ice until above are defined.

Feedback is welcome.

BitRate27:
Thanks for looking into this PR further. While I agree that the OBS Studio API should provide this information: is the source visible in program, preview or both, the solution in this PR does address the challenges you mentioned.  It does this by not using the hide/show/activate/deactivate APIs and instead looks at the current program scene and preview scene, specifically looking to see if the source is visible in the scene and setting the tallies appropriately. It knows nothing, nor does it care, about multiview or other displays of sources. I encourage you to build and run this PR to see that it does indeed address the challenges you mentioned, by basically not using those existing APIs.
Thanks for finding the OBS API issue! I could not find one and will pursue a change in OBS through that issue. As I said, previously, once that API shows up, we can remove the workaround code in this PR which searches the scenes for the source.
As for the seriousness of the issue, I found the tally light issue when I first started using obs-ndi. It was confusing for me and really distracted me from finding a good production operation plan. I just decided not to use tallys for my cameras. Instead of submitting a new issue, I found the existing one.

BitRate27:
I upticked the suggestion for adding the API to OBS. Only 2 upticks! Please consider upticking also: https://ideas.obsproject.com/posts/1271/add-obs_source_inpreview-to-sdk

BitRate27:
> ie.: should a source where and NDI Filter (output) is used also consider updating the Tally ? - depends
 
I would say No. The send API in NDI manages the reporting of tally to any viewers via meta data. We dont need to do anything.

BitRate27:
This PR is superceeded by [PR 1347](https://github.com/DistroAV/DistroAV/pull/1347)

PR #1084, opened 2024-08-29
Check for ndiLib when destroying receiver and framesync

BitRate27:
The new 100ms delay after check for no_of_connections == 0 can mean the thread loop will exit after the ndiLib has been destroyed in module unload. Adding a check for a valid ndiLib link before calling destroy methods will keep shutdown from crashing.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -776,12 +776,12 @@ void *ndi_source_thread(void *data)
 	//
 
 	if (ndi_frame_sync) {
-		ndiLib->framesync_destroy(ndi_frame_sync);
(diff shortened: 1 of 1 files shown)


Trouffman:
Great! This was something I was experiencing as well recently, even added a comment about it in the code.

Looking to see if this can be added in the next release too.

Trouffman (approved):
All sounds good to me.
Also tested this code in the Visibility Behavior Update Branch : https://github.com/DistroAV/DistroAV/pull/1074

BitRate27 (commented):
Looks good! Thanks for adding.

PR #1092, opened 2024-09-09
Recognize ndi_source_name change

BitRate27:
When the source name changes, the ndi receiver would not get reset. This was because the ndi_source_name is a char * in the config and the code was in effect comparing with itself the second time through the update code. Fixed by changing ndi_source_name to a std::string.


Trouffman:
This does not seem to solve the issue.

Trouffman:
Seems the breaking change appeared in : https://github.com/DistroAV/DistroAV/commit/591a58e454bd174ad91e6280c002e8c8c12dd127


Note : It will usually work the first time you cahnge the source name, but not the second time (because the ndi_source_name is empty riht after start (seem wrong as well)

Trouffman:
This was fixed in the other pr for #1074

PR #1137, opened 2024-11-08
Add a log entry with the exact path to the old OBS-NDI plugin file

Trouffman:
This adds a log entry with the exact path to the old OBS-NDI plugin file to help with diagnosing support request for issues like "DistroAV still detect OBS-NDI".

The message to look for in the OBS log is : 

```
DistroAV] is_module_found: `obs-ndi` found at `/Users/[USER]/Library/Application Support/obs-studio/plugins/obs-ndi.plugin/Contents/MacOS/obs-ndi` and '/Users/[USER]/Library/Application Support/obs-studio/plugins/obs-ndi.plugin/Contents/Resources'
[DistroAV] obs_module_load: OBS-NDI is detected and needs to be uninstalled before DistroAV will load.
```

This should then help in the Troubleshooting guide as well : https://github.com/DistroAV/DistroAV/wiki/OBS%E2%80%90NDI-Is-Now-DistroAV

### DIFF
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -188,21 +188,15 @@ bool is_module_found(const char *module_name)
 		[](void *param, const struct obs_module_info2 *module_info) {
 			struct find_module_data *data_ =
 				(struct find_module_data *)param;
-			if (stricmp(data_->target_name, module_info->name) ==
+			if (strcmp(data_->target_name, module_info->name) ==
 			    0) {
-				obs_log(LOG_WARNING,
+				obs_log(LOG_INFO,
 					"is_module_found: `%s` found at `%s`",
 					module_info->name,
 					module_info->bin_path);
-				obs_log(LOG_INFO,
-					"is_module_found: Found module_info->name == `%s`",
-					module_info->name);
-				obs_log(LOG_EBUG,
-				     "is_module_found: module_info->bin_path=`%s`",
-				     module_info->bin_path);
 				obs_log(LOG_DEBUG,
-				     "is_module_found: module_info->data_path=`%s`",
-				     module_info->data_path);
+					"is_module_found: module_info->data_path=`%s`",
+					module_info->data_path);
 				data_->found = true;
 			}
 		},
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -194,17 +194,15 @@ bool is_module_found(const char *module_name)
 					"is_module_found: `%s` found at `%s`",
 					module_info->name,
 					module_info->bin_path);
-#if 0
 				obs_log(LOG_INFO,
 					"is_module_found: Found module_info->name == `%s`",
 					module_info->name);
-				obs_log(LOG_INFO,
+				obs_log(LOG_EBUG,
 				     "is_module_found: module_info->bin_path=`%s`",
 				     module_info->bin_path);
-				obs_log(LOG_INFO,
+				obs_log(LOG_DEBUG,
 				     "is_module_found: module_info->data_path=`%s`",
 				     module_info->data_path);
-#endif
 				data_->found = true;
 			}
 		},
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -188,10 +188,10 @@ bool is_module_found(const char *module_name)
 		[](void *param, const struct obs_module_info2 *module_info) {
 			struct find_module_data *data_ =
 				(struct find_module_data *)param;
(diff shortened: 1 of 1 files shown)


src/plugin-main.cpp
```diff
@@ -190,13 +190,18 @@ bool is_module_found(const char *module_name)
 				(struct find_module_data *)param;
 			if (strcmp(data_->target_name, module_info->name) ==
 			    0) {
+				obs_log(LOG_WARNING,
+					"is_module_found: `%s` found at `%s` and '%s'",
```
paulpv, line 194: invalid number of arguments; 3 expected, 2 actual

paulpv:
I have a similar change for this that I haven't committed yet:
image

I also want to change the installer display name (from `distroav` to `DistroAV`):
image

src/plugin-main.cpp
```diff
@@ -188,19 +188,21 @@ bool is_module_found(const char *module_name)
 		[](void *param, const struct obs_module_info2 *module_info) {
 			struct find_module_data *data_ =
 				(struct find_module_data *)param;
-			if (strcmp(data_->target_name, module_info->name) ==
+			if (stricmp(data_->target_name, module_info->name) ==
```
paulpv, line 191: Oh, I didn't really mean for you to use my code; I liked your code better.
TIL that ANSI C does not define `stricmp`!
Supposedly there is `strcasecmp`, but supposedly that is also not guaranteed.
Maybe just revert back to `strcmp`?

src/plugin-main.cpp
```diff
@@ -188,19 +188,21 @@ bool is_module_found(const char *module_name)
...
 			    0) {
+				obs_log(LOG_WARNING,
```
paulpv, line 193: LOG_INFO

src/plugin-main.cpp
```diff
@@ -188,19 +188,21 @@ bool is_module_found(const char *module_name)
...
+					"is_module_found: `%s` found at `%s`",
+					module_info->name,
+					module_info->bin_path);
 				obs_log(LOG_INFO,
```
paulpv, line 197: Remove

src/plugin-main.cpp
```diff
@@ -188,19 +188,21 @@ bool is_module_found(const char *module_name)
...
 					"is_module_found: Found module_info->name == `%s`",
 					module_info->name);
-#if 0
-				obs_log(LOG_INFO,
+				obs_log(LOG_EBUG,
```
paulpv, line 200: Remove

Trouffman:
Latest commit teste and output expected result. 
image

paulpv:
> Latest commit teste and output expected result. image

Yeah, that is what I am a little worried about.
I am not **very** worried, but I would prefer to not just post people's names into OBS logs that are on the internet.
I think that might even violate some privacy laws. :/
Software is complicated nowadays beyond just the code. :/

Did you run this with any command-line arguments (ex: `--distroav-debug`), or none?

paulpv:
One option is to change the `LOG_INFO` to `LOG_DEBUG`, but then that would require someone to enter `--distroav-debug` on the command-line, which very few people will do.

Another option is to redact the username, but redacting the path almost defeats the purpose of printing out the path.

Trouffman:
> > Latest commit teste and output expected result. image
> 
> Yeah, that is what I am a little worried about. I am not **very** worried, but I would prefer to not just post people's names into OBS logs that are on the internet. I think that might even violate some privacy laws. :/ Software is complicated nowadays beyond just the code. :/
> 
> Did you run this with any command-line arguments (ex: `--distroav-debug`), or none?

This was run as-is without the debug argument.

I am not keen on having the user folder as well being displayed.

paulpv:
> I am not keen on having the user folder as well being displayed.

We could just have it redacted to say something like:
```
/Users/$USER/Library/Application\ Support/obs-studio/plugins/obs-ndi.plugin/
```

That seems both safe and still usable.

Trouffman:
Added a filter for privacy. 
Not ideal.

Tested on MacOS :

`is_module_found: `obs-ndi` found at `[Redacted User Home Path]/Library/Application Support/obs-studio/plugins/obs-ndi.plugin/Contents/MacOS/obs-ndi`

Using DEBUG log-level will still display the full path.

Trouffman:
@paulpv Please review the latest changes to add some privacy to the path.

Trouffman:
This error-tweak can (should) be removed in future version after migration from pre distro-av to Distro-av is with very high adoption.

Issue #1147, opened 2024-11-18
[Bug]: OBS crash when turning on main output after choosing an unsupported color format


BitRate27:
### Operating System Version

macOS Monterrey, Windows 10

### OBS Version

31.0.0

### OBS Installation Method

exe, pkg

### DistroAV Version

6.0.0

### DistroAV Installation Method

exe, pkg

### NDI Version

6.0.1

### Describe the bug

When you turn on main output in DistroAV NDI settings after you have chosen an unsupported color format, OBS will crash inside OBS due to a problem with the output->video. 

### Steps to reproduce the problem

1. In Tools->DistroAV NDI settings, turn OFF Main output. Select OK.
2. In Settings->Advanced, for Video Color Format choose: P010. Select OK.
3. In Tools->DistroAV NDI settings, turn ON Main output. Select OK.
4. In log file, should see: unsupported pixel format n.
5. In Tools->DistroAV NDI settings, turn OFF Main output. Select OK.
6. In Settings->Advanced, for Video Color Format choose NV12. Select OK.
7. In Tools->DistroAV NDI settings, turn ON Main output. Select OK.
Will crash OBS on Windows and macOS at this point.

### Expected behavior

The Main output should be turned on and OBS should be outputting the Main program output to NDI.

### Additional context

The crash is here when looking at video->parent. Normally parent is 0000x0 but if the main output creation is aborted due to an unsupported pixel format, the next time a main output is created, the parent is non-null and causes a crash.

From: libobs/media-io/video-io.c:
```
static const video_t *get_const_root(const video_t *video)
{
	while (video->parent)
		video = video->parent;
	return video;
}
```

PR #1148, opened 2024-11-18
Show user why NDI output failed

BitRate27:
When the user selects a color format of P010, I010, P216 or P416, the output will not be started because the NDI SDK does not support HDR color formats. The only indication this is a problem is by looking for "unsupported pixel format" in the log file. To make it easier for the user to identify the problem, we will now report the last error in the DistroAV NDI Settings dialog for the main output. 

image

This message will be seen when the user goes back to the settings dialog after noticing there is no NDI output.

This does not happen for preview or dedicated NDI output from a filter, so no changes were needed for those outputs.

In order to test and utilize this new message, this PR also fixes https://github.com/DistroAV/DistroAV/issues/1147 where OBS crashes after you have tried to output an unsupported format. 

The problem existed because an output would be created even when no main output was selected. When the user went to change the color format, in the OBS Settings->Advanced dialog, this would invalidate the video output in DistroAV NDI Settings. Then when main output was turned back on, OBS would crash because output->video is corrupted. 

This makes the changes to main-output look big but it was mainly removing an indented if condition. In summary we always call main_output_deinit in main_output_init, and then if it is not enabled, we return, otherwise we create the output and start it every time it is enabled. 

Tested on Windows by testing that all formats gave the correct message if unsupported, or blank if supported.


RayneYoruka:
I like the idea of this being added, please @paulpv or @Trouffman check it

src/main-output.cpp
```diff
@@ -83,13 +83,20 @@ void main_output_start()
 			obs_log(LOG_INFO,
 				"main_output_start: successfully started NDI main output '%s'",
 				QT_TO_UTF8(context.ndi_name));
+			Config::Current()->OutputLastError = QString("");
 		} else {
 			auto error = obs_output_get_last_error(context.output);
+			Config::Current()->OutputLastError =
+				"Last Error: Unsupported color format " +
```
Trouffman, line 90: Can we localize this error message? Let's avoid this to be hard coded if visible to end-users.

Trouffman:
Thanks for the extensive PR.
Is there a way that it could be displayed without being set in the config file?

I see we could re-use the pixel format detection as well if one day we can use the advanced SDK.

BitRate27:
I think it is possible to add a method to set the last error to the OutputSetting dialog and display that string instead when the dialog is displayed. I can redo that part.

BitRate27:
I am looking into a way to keep the user from turning on main output if the color format not supported.

BitRate27:
After this latest commit https://github.com/DistroAV/DistroAV/commit/ef7e0ec35f1bcd56e3209b510668dbb080f7b6f8 , the main output box is only enabled if the pixel format of the main output is supported by the NDI SDK. If it is not supported, the main output box is disabled and an error message appears showing why main output cannot be chosen.

Trouffman:
Tested on MacOS. Behave as expected with this PR. Including with changing some settings in the config files too behave properly.

Trouffman:
I just found a bug while investigating another behavior.

Documented bug at : https://github.com/DistroAV/DistroAV/issues/1180

Trouffman:
@BitRate27 I merged from the latest master, please review to confirm that this still has the intended purpose.

BitRate27:
> With the build files at : https://github.com/DistroAV/DistroAV/actions/runs/13763251660/job/38483860839?pr=1148 On MacOS the Output option stay greayed out, no error message. even when using compatible format.

I see the same thing, with your change. Hmm...

BitRate27:
I have got a fix for this, however, while testing I found out we are no longer outputting the preview scene. I updated the https://github.com/DistroAV/DistroAV/pull/1198 PR to indicate the issue found, but will most likely fix that problem in this PR as it is kind of tied to the logic of trying to see if we can output Main.

Trouffman:
@BitRate27 Any chance this could be fixed asap for the Preview scene?
As this is creating a regression, it's like a blocking to move ahead with more release.

BitRate27:
I cannot get preview to work in DistroAV 6.0.0 and the latest OBS master. Have you been able to duplicate? Looks like the regression happened long before these changes.

BitRate27:
I verified my latest changes to the "show user why output failed" PR address the original issue along with not allowing main output selection if bad pixel format. Also, verified main output still works when pixel format is supported. Was not able to verify preview output changes due to overall problem with Preview which seem unrelated to this PR.

Trouffman:
@BitRate27 please review the latest changes on this following the conflict resolution.

Trouffman (commented):
Issue found : This is "Stopping" the Main Output when opening the DistroAV Settings and is not re-created if "cancelled".

Expected Behavior : The stream should NOT stop until there is a need for:
- Stopping (aka not enabled)
- Name or group name changed
All above only after the form has been submitted.

Error log shows : `info: Output 'NDI Main Output': stopping`

Trouffman:
This is what happen when opening the NDI Settings dialog box :

`08:28:52.056: [distroav] +main_output_init()
08:28:52.057: [distroav] +main_output_deinit()
08:28:52.057: [distroav] main_output_deinit: NDI Main Output is not initialized
08:28:52.057: [distroav] -main_output_deinit()
08:28:52.057: [distroav] main_output_init: creating NDI Main Output 'OBS-PGM'
08:28:52.057: [distroav] Creating NDI Main Output : 'OBS-PGM'
08:28:52.057: [distroav] +ndi_output_getdefaults()
08:28:52.057: [distroav] -ndi_output_getdefaults()
08:28:52.057: [distroav] +ndi_output_create(name='OBS-PGM', groups='', ...)
08:28:52.057: [distroav] ndi_output_update(name='OBS-PGM', groups='', ...)
08:28:52.057: [distroav] NDI Output Updated. 'OBS-PGM'
08:28:52.057: [distroav] ndi_output_update(name='OBS-PGM', groups='', uses_video='true', uses_audio='true')
08:28:52.057: [distroav] -ndi_output_create(name='OBS-PGM', groups='', ...)
08:28:52.057: [distroav] main_output_init: created NDI Main Output 'OBS-PGM'
08:28:52.057: [distroav] +main_output_start()
08:28:52.057: [distroav] main_output_start: starting NDI Main Output 'OBS-PGM'
08:28:52.057: [distroav] +ndi_output_start(name='OBS-PGM', groups='', ...)
08:28:52.058: [distroav] +on_main_output_started()
08:28:52.058: [distroav] -on_main_output_started()
08:28:52.058: [distroav] NDI Main Output started
08:28:52.058: [distroav] 'OBS-PGM' ndi_output_start: ndi output started
08:28:52.058: [distroav] -ndi_output_start(name='OBS-PGM', groups=''...)
08:28:52.058: [distroav] main_output_start: successfully started NDI Main Output 'OBS-PGM'
08:28:52.058: [distroav] -main_output_start()
08:28:52.058: [distroav] -main_output_init()
08:28:52.058: [distroav] +main_output_deinit()
08:28:52.058: [distroav] +main_output_stop()
08:28:52.058: [distroav] main_output_stop: stopping NDI Main Output 'OBS-PGM'
08:28:52.058: [distroav] +ndi_output_stop(name='OBS-PGM', groups='', ...)
08:28:52.058: Output 'NDI Main Output': stopping
08:28:52.058: Output 'NDI Main Output': Total frames output: 0
08:28:52.058: Output 'NDI Main Output': Total drawn frames: 0
08:28:52.059: [distroav] +on_main_output_stopped()
08:28:52.059: [distroav] -on_main_output_stopped()
08:28:52.059: [distroav] NDI Main Output stopped
08:28:52.059: [distroav] ndi_output_stop: +ndiLib->send_destroy(o->ndi_sender)
08:28:52.063: [distroav] ndi_output_stop: -ndiLib->send_destroy(o->ndi_sender)
08:28:52.063: [distroav] NDI Output Stopped. 'OBS-PGM'
08:28:52.063: [distroav] -ndi_output_stop(name='OBS-PGM', groups='', ...)
08:28:52.063: [distroav] main_output_stop: successfully stopped NDI Main Output 'OBS-PGM'
08:28:52.063: [distroav] -main_output_stop()
08:28:52.063: [distroav] main_output_deinit: releasing NDI Main Output 'OBS-PGM'
08:28:52.063: [distroav] +ndi_output_destroy(name='OBS-PGM', groups='', ...)
08:28:52.063: [distroav] -ndi_output_destroy(name='OBS-PGM', groups='', ...)
08:28:52.063: [distroav] main_output_deinit: successfully released NDI Main Output ''
08:28:52.063: [distroav] -main_output_deinit()
`

This is because of the logic for `main_output_is_supported()` that will "test" by creating the output. But this conflict with the "on-going" output.

This should be triggered only when loading a new profile / starting OBS and not everytime the NDI Settings are opened.

BitRate27:
Thanks for looking at this. You brought up some good issues with this implementation. Clearly I need to rethink the logic here. I think we need to initialize a test output instead of using the main output. Not sure how to only show the "last error" row in the dialog if no error. Suggestions welcome. 

Here are the scenarios:
1. New profile loaded with main output OFF and format NOT supported
    - Turn off main output
    - Main output: not started
    - Dialog: Main output group disabled, not checked and error shown
2. New profile loaded with main output OFF and format supported
    - Turn off main output
    - Main output: not started
    - Dialog: Main output group enabled, not checked and no error shown
        - OK and main output checked -> Main output started
        - OK and main output not checked -> Main output not started
        - Cancel -> Main output not started
3. New profile loaded with main output ON and format supported
    - Turn off main output
    - Main output: started
    - Dialog: Main output group enabled, checked and no error shown    
4. User changes to a format NOT supported
    - Dialog: Main output group disabled, not checked and error shown
    - OK/Cancel: Main output not started
5. User changes to a format supported
    - Dialog: Main output group enabled, not checked and no error shown
        - OK and main output checked -> Main output started
        - OK and main output not checked -> Main output not started
        - Cancel -> Main output not started

BitRate27:
Update: was able to alter the main_output_is_supported to check without using main output. Also, figured out how to set the message height to zero if no error message. Only thing left is to work on the LOG_INFO and LOG_DEBUG messages. All tests in the scenarios above passed.

src/forms/output-settings.cpp
```diff
@@ -226,7 +230,14 @@ void OutputSettings::showEvent(QShowEvent *)
 	ui->mainOutputGroupBox->setChecked(config->OutputEnabled);
 	ui->mainOutputName->setText(config->OutputName);
 	ui->mainOutputGroups->setText(config->OutputGroups);
-	ui->mainOutputLastError->setText(main_output_last_error());
+
+	auto lastError = main_output_last_error();
+	ui->mainOutputLastError->setText(lastError);
+	if (lastError.isEmpty()) {
+		ui->mainOutputLastError->setFixedHeight(0); // don't was dialog space
```
Trouffman, line 237: Typo to correct.

src/main-output.cpp
```diff
@@ -100,18 +100,44 @@ void main_output_start()
...
+	}
+	
 	return is_supported;
 }
 
 void main_output_deinit()
 {
-	obs_log(LOG_DEBUG, "+main_output_deinit()");
 	if (context.output) {
+		obs_log(LOG_DEBUG, "+main_output_deinit()");
```
Trouffman, line 138: This would be confusing as the "debug entry to main_output_deinit()" will only show when it is doing something and not "everytime it is called"

obvious path would be : `obs_log(LOG_DEBUG, "+main_output_deinit()"` right after entering the function.

There is a log entry when we get into main_output_stop() so no need to add one here.

BitRate27: I moved it back outside the if condition. Was mainly trying to remove the log message when it says it failed to initialize when it hadn't been initialized before. Guess I went too far in cleaning it up.

Trouffman:
Tested on Windows 11+ OBS 31 after the latest merge :
- The default output does not work once  a non compatible profile is used, same when going back to an expected to work profile.

Seem our logic here is not polished enough.

BitRate27:
Looks like the one scenario I didnt list above. Will check it out. Thanks!

Trouffman:
Tests results on Windows 11

Scenarios : 

    New profile loaded with main output OFF and format NOT supported
        Turn off main output
        Main output: not started
        Dialog: Main output group disabled, not checked and error shown

-> OK

    New profile loaded with main output OFF and format supported
        Turn off main output
        Main output: not started
        Dialog: Main output group enabled, not checked and no error shown
            OK and main output checked -> Main output started
            OK and main output not checked -> Main output not started
            Cancel -> Main output not started
-> OK

    New profile loaded with main output ON and format supported
        Turn off main output
        Main output: started
        Dialog: Main output group enabled, checked and no error shown

-> OK

    User changes to a format NOT supported
        Dialog: Main output group disabled, not checked and error shown
        OK/Cancel: Main output not started

-> OK

    User changes to a format supported
        Dialog: Main output group enabled, not checked and no error shown
            OK and main output checked -> Main output started
            OK and main output not checked -> Main output not started
            Cancel -> Main output not started

->ok

BitRate27:
```
New profile loaded with main output ON and format NOT-supported
    Turn off main output (?)
    Main output is still checked (as it was loaded as "active") BUT this is not running.
    User change Settings to a Supported pixel format
    The Main Output should restart, if the format is supported 
```
This can happen if the profile was saved before this code was implemented. Valid case!
(?) How do you turn off main output with this new code if the format is not supported? The checkbox should not be sensitive.

Trouffman:
I could reproduce on Mac with this :

- Open Open with supported profile & Main Output ON
- Load non-compliant profile
- Behavior (expected) : Main output OFF, greyed out option, box is still ticked.
- Go to Settings : change to a valid pixel format
Behavior : Main output OFF, option available, box is still ticked.
> To re-enable, the Main Output must be disabled, saved, then re-enabled.

I think this need to be clear-er on the approach we take:
1. Should the option be "disabled" if a Pixel format is not supported
or
2. Should we build a (more complex) logic to re-enable the Main Output when the pixel format is changed AND the box ticked.


Option 1. would touch the configuration file
Option 2. would need to have a check for specifically this change as we do NOT want to stop the main output everytime we go in the OBS settings.

Trouffman:
@BitRate27 This PR is proving to be more complex than expected.

While doing the review I found a couple extra comment (not blocking but would be good to have at some point)

1. Should we call `main_output_is_supported()` everytime we plan to use `main_output_init` ?

ie. : in `plugin-main.cpp` line 299


2. `main_output_is_supported()` is calling the `obs_output_create` with an "empty" ndi_name.

This did not create issues when testing, hopefully this would stay this way but having a clear "testing name" might help to identify in error log as well.

We could put it in a group called "DistroAV Config" so this does not even show up anywhere when testing.

3. Following on point 2. above : Giving the output a unique name might avoid weird issues too (even tho the OBS documentation says that it will make it unique).


This PR brings a lot of value as it will help for the support and anyone using HDR pixel format.

BitRate27:
Disabling main output in the config if the format is not supported, keeps the toggle in sync with the main output status in OBS.

Also, moving the check for support into the main_output_init allows us to turn it off if not supported and enabled.

Finally, with the output enabled in sync with the OBS main output state, main_output_init is only called when settings of main output are changed. 

I verified the original scenarios still work, and the new scenario performs as it should. Also added a LOG_INFO message to let the user know the output was disabled because the format is not supported.

Trouffman:
Great! Thank you.

I updated the INFO to a WARNING and attributed ID 426 to easily flag it too.

PR #1152, opened 2024-11-20
Use new obs_frontend_get_*_config APIs for OBS 31

### LINKED ISSUE #1121: [Bug]: OBS log reports `obs_frontend_get_global_config is deprecated.`
### Operating System Version

Windows 11

### CPU

_No response_

### GPU

_No response_

### NDI Devices

_No response_

### OBS Version

30.2.3

### OBS Installation Method

exe

### DistroAV Version

6.0.0

### DistroAV Installation Method

exe

### [Extra] Installation Steps

_No response_

### OBS Log [URL]

https://obsproject.com/logs/GQ5BKds8dx4VSy9e

### NDI Version

6.0.1

### Describe the bug

The OBS log reports the following when initializing DistroAV:
```
01:40:13.807: DEPRECATION: obs_frontend_get_global_config is deprecated. Read from global or user configuration explicitly instead.
01:40:13.807: DEPRECATION: obs_frontend_get_global_config is deprecated. Read from global or user configuration explicitly instead.
```

The code needs to do what that says: stop using `obs_frontend_get_global_config` and read from global/user config explicitly.

The code calls it here:
https://github.com/search?q=repo%3ADistroAV%2FDistroAV%20obs_frontend_get_global_config&type=code
https://github.com/DistroAV/DistroAV/blob/7126fc4bc173187010c0d0e0f8f5f38754bf5f09/src/obs-support/obs-app.hpp#L34

Should be a simple fix.

kaechele:
This properly splits settings for this plugin into the new split configuration stores for app and user settings.

All user-editable settings go into the user store, all functional, non-user facing settings go in app store (e.g. timestamp of last update check).

On OBS versions prior to 31 there is no change in behaviour and all settings continue to be read from and stored to the global store only.

Fixes #1121


kaechele:
Took a first stab at a migration functionality. Works fine in my testing here, but I'd certainly appreciate some more testing.
It should:
* migrate the settings that are now user settings from the global scope to the user scope
* delete migrated settings from the global scope

If everything worked your original settings from `global.ini` should now be in `user.ini` and only app scope settings should remain in `global.ini`.

Trouffman:
Thank you for this change proposal.
Note for testers : This should be tested on OBS 31 RC1 or more recent for pre-validating the changes.

Trouffman:
Merging things from the updated OBS Plugin Template for OBS 31 "base".

Due to limited feedback from PR submitter, will try to continue from this to a workable solution.

Trouffman:
Update to some concerns in this PR : 

PARAM_SKIP_UPDATE_VERSION
Can be set & influenced by the user, the decision is to make it a "User" config

The other elements are not influenced / presented to the User and will be in the App Config:
PARAM_LAST_UPDATE_CHECK
PARAM_MIN_AUTO_UPDATE_CHECK_INTERVAL_SECONDS
PARAM_AUTO_CHECK_FOR_UPDATES


This is based on the commit infos :obsproject/obs-studio@2635cf3

    The app config instance contains system-wide settings that mainly
    concern the working of the app itself, whereas the user config instance
    contains settings actually exposed to the user for the configuration.

obsproject/obs-studio#11055

Trouffman:
While digging to update the plugin i foudn out that the check with LIBOBS_API_MAJOR_VER will only check at compile time and not Runtime.

if we want to make the plugin backward compatible we will need to either expand on [obs_get_version](https://docs.obsproject.com/reference-core#c.obs_get_version) or simply drop this complexity and make OBS 31 a requirement.

Trouffman:
As we made OBS 31 a full requirement for DistroAV 6.1 this can be revisited and simplified.

PR #1158, opened 2024-12-05
Remove obs-ndi on macos after install of DistroAV

BitRate27:
Execute a postinstall script on MacOS to remove the obs-ndi plugin from OBS in the users Library and the system library if permissible. Will also remove the script and the scripts directory afterwards.

This script was added because many users got blocked from using DistroAV when they failed to follow the instructions for migrating to the new plugin name correctly.


BitRate27:
Since we are updating the install scripts, I also added the opening of the firewall for OBS on mac and windows.
I was able to test the install on windows, but was only able to test the postinstall script on mac by manually invoking it. 
To test I did the following on on both OSs:

1. Output NDI from another PC (make sure firewall is off)
2. Turn on the firewall on the PC being tested
3. Run OBS
4. Create a NDI source
5. Verify the other computer cannot be selected as a source
6. Exit OBS
7. Install the DistroAV plugin
8. Run OBS
9. Select the new NDI source 
10. Select Properties
11. Should be able to see, select and preview the NDI source from the other computer now

There may have been discussions on why we shouldn't do this during install. If this is not the way we want to go, I can gladly remove these commits from the PR.

Trouffman:
Tested the MacOS fix, and there is no residual postinstall script.

Trouffman (approved):

Issue #1164, opened 2024-12-16
[Bug]: VLC source  NDI Filter  out of sync


lifei7408:
### Is there an existing issue for this?

- [X] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [X] I have checked this bug still exist on the latest available version of DistroAV
- [X] I have checked that my OS / OBS / NDI versions are supported

### OBS Log [URL] - Important!

https://obsproject.com/logs/sW8FOLSphDHvgWYJ

### Issue reported - Current Behavior

When using the VLC source, the sound and video output by the NDI Filter are out of sync, and the audio leads the video by about 1 second.

### Expected Behavior

Audio and video synchronization

### Steps To Reproduce

Install plugin DistroAV.
using the VLC source.
Set up NDI Filter.
play meidia in VLC source.
Use ndi monitr to receive

### Environment

```markdown
- OS:windows 10 22H2
- OBS:31.0.0
- DistroAV: 6.0.0
- NDI:6.0.1
```

### Pre-Check for Bug Report

- [X] I am able to reproduce the bug following the steps provided above
- [X] I have confirmed this bug exist on more than one device/computer
- [X] I have tested using a new scene collection to reproduce the issue
- [X] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

BitRate27:
If you record the VLC video source in OBS, do you still see the delay?

lifei7408:
> If you record the VLC video source in OBS, do you still see the delay?

no delay！

and  distroAV NDI  Main output  no delay too

RayneYoruka:
Since this is solved lets close this issue x)

Trouffman:
This seems to be a performance limitation on your system and mostly a playback performance issue.

Make sure to check that VLC use hardware acceleration as well.

As this is not related to the plugin we will close this issue.

BitRate27:
I have been trying to duplicate this issue. I setup two sources, one as a media player source, and one as a VLC source. I chose the same file. On each source I added a dedicated NDI output filter.  For media player, I saw a 1/2 second sync issue between the audio and video when viewing the dedicated output in NDI Studio Monitor. For the VLC, I got no video or audio output from the dedicated filter.

I would be curious if [lifei7408](https://github.com/lifei7408) sees the same delay when using a media player source. 

[...]

PR #1172, opened 2024-12-24
Synchronize VLC audio with video in dedicated NDI output filter

BitRate27:
This is to fix issue [1164](https://github.com/DistroAV/DistroAV/issues/1164).

The video output by a dedicated NDI output filter when added to a VLC source showed a 1.777 second delay compared to the audio. 

This was happening because of two problems in the ndi-filter.cpp code:
1. The timestamp of the video was being overridden by the current time in ndi_filter_offscreen_render. 
2. The NDI library was not told to synchronize the audio and video when the sender was created.

It appears as though the problem was unique to VLC. When the filter was put on a Media Source with the same file, everything was in sync. VLC is delivering video to OBS at a delayed offset and no effort was being done to resync the audio and video in the filter. In fact the filter was actually overwriting the video time stamps making it impossible to synchronize on output.

Another by product of this bug was with 60fps VLC sources. Audio was only being sent through randomly, mostly not at all. So with this fix, 60fps video is output with the correct sound, and in sync.

### DIFF
--- a/src/ndi-filter.cpp
+++ b/src/ndi-filter.cpp
@@ -200,7 +200,7 @@ void ndi_filter_offscreen_render(void *data, uint32_t, uint32_t)
 
 		video_frame output_frame;
 		if (video_output_lock_frame(f->video_output, &output_frame, 1,
-					    os_gettime_ns())) {
+					    obs_get_video_frame_time())) {
 			if (f->video_data) {
 				gs_stagesurface_unmap(f->stagesurface);
 				f->video_data = nullptr;
@@ -241,8 +241,8 @@ void ndi_filter_update(void *data, obs_data_t *settings)
 		send_desc.p_groups = groups;
 	else
 		send_desc.p_groups = nullptr;
-	send_desc.clock_video = false;
-	send_desc.clock_audio = false;
+	send_desc.clock_video = true;
+	send_desc.clock_audio = true;
 
 	if (!f->is_audioonly) {
 		pthread_mutex_lock(&f->ndi_sender_video_mutex);


Trouffman:
Did you test with other filter (without VLC) does this impact the Sync as well?

My understanding for the review is that :
- The Sync for the filter was done with the wrong time stamp (general time instead of the frame time)
- No sync for filter (This reminds me a ticket( very old one - but could not find it back.)

BitRate27:
I tested that the media source still works correctly with the fixed dedicated ndi filter.

BitRate27:
@Trouffman brought up that the obs_get_video_frame_time() method is not part of the official OBS API. This is also true for all of the video_* methods called in the offscreen_render callback. This non-supported video-io API is used for dedicated filter output of NDI so that OBS does the format conversion and we can output the NDI in a common format, the one supported by OBS main output. If we don't use that hidden API, then we need to do the conversion of whatever source the filter is on, to a format supported by NDI. Also, by tagging on to the main OBS output, we are taking advantage of the syncing of audio and video OBS is doing. So in addition to format conversion, we need to sync the audio and video.

Trouffman:
As @BitRate27 pointed out, most of the code and function used for the filter, are made on "deep" OBS function ( not the one for the documented API).

This means, we can only "fix" things this way.
Ultimately we need to clean the code but this is not the goal of this PR.

Left to do : test this PR with other source (than VLC) to not break something else.

Also need to check if this would solve some of the issues mentioned at : https://github.com/norihiro/obs-async-audio-filter/tree/main#recommended-settings

BitRate27:
My Windows testing continues...I found two other sources, in addition to VLC, which output audio and video: Media Source and NDI Source.

Using these files (60fps and 30fps) as input to the Media and VLC sources:

https://github.com/user-attachments/assets/e3bc8763-1c51-4fc5-b621-2cd315214f4c
https://github.com/user-attachments/assets/f9a636dd-b7fa-47b6-bd01-b6baf3712daf

I visually monitored the dedicated NDI filter output in Studio Monitor, and recorded the main output to a file, where I viewed the recording in Windows Media Player. In all cases: 30/60 fps x 3 sources x NDI and recorded = 12 combinations. The video played in sync visually in all cases.

Testing the recording made sure locking the frame with video_frame_time on the main output did not change the behavior of the main output.

BitRate27:
Looking in detail at the delay of the video for the main output for all 3 sources before (Orig) and after (Fix):
image

This shows that with the fix, the NDI source video delay is the same, the VLC video delay is about 30ms less, and the Media video is about 70ms more delayed. The shortest delay before the fix was about 130ms for Media source.

To give a since of size here, the VLC NDI output had a 1777ms delay, with this fix it is down to less than 200ms.

PR #1175, opened 2024-12-30
Upgrade to use audio_frame v3 instead of deprecated v2

Trouffman:
Add logging for timestamps & Timecode
Force timecode to be provided by the NDI lib instead of hardcoded in plugin
Unifiy usage accross recv/sender
Update filter audio to v3

Force FrameSync to use Audio_frame v3 instead of v2.

This is not really supposed to fix any specific bugs, but should help to keep the codebase up to date.

I have tested locally, but I do not use the NDI Filter that much so could be missing some things.

Things to test :
- NDI source with Framesync ON. (now using audio_frame3 instead of audio_frame2) - report if no working.
- NDI Filter (audio only) - If not working please report.
- Extra TS/TC debugging (Verbose mode only) - Report any inconsistencies


Trouffman:
Looking at the historical changes for this piece of code and how the "video_frame.timecode" is set I found that:

- It used to be the NDI Lib synthetise options around here : https://github.com/DistroAV/DistroAV/pull/13
- This was also suggested to be a fix at : https://github.com/DistroAV/DistroAV/pull/564

Great summary PR #564 

To confirm : why was this reverted...

RayneYoruka:
> I have tested locally, but I do not use the NDI Filter that much so could be missing some things.

Since I run so many NDI filter sources I can test it today.

BitRate27:
Minor suggestion is to remove the audio_frame2 variable from the ndi_source_thread since it is no longer used. Would also be nice to change the video_frame2 and audio_frame3 variables to video_frame and audio_frame.

Trouffman:
This should be "complete". I haven't experienced any issue with it since last commit. Used on Mac & windows.

RayneYoruka:
So far it's been fine. Should push IMO

PR #1186, opened 2025-01-18
Wait for all NDI sources before filling out source list

BitRate27:
Sometimes NDI sources do not appear in the NDI sources list in the properties dialog for DistroAV. Overtime they will eventually appear. Newly added sources from newly attached devices seem to be the ones that are missing. It has been hard to duplicate a repeatable sequence.

This PR changes the way we use the find API in NDI SDK. Instead of creating the find when the source is created, we create the find everytime the user opens the properties dialog. Also, we loop through the find until we find no more sources, then fill in the list.

This change is based on the examples provided in the NDI Advanced SDK.


Trouffman (commented):
Thanks for the PR. Only challenge I see is that it delays the display of the dialog box and will impact user-experience.

Delays will ALWAYS be of 1sec then 1sec extra everytime there is a source being discovered.

Lookign to find a way to update while not impacting the loading time of the dialog box.

BitRate27:
Here is what it says about this in the NDI SDK:

_It is important to understand that mDNS discovery might take some time to locate all network sources. This means that an ‘early’ return to NDIlib_find_get_current_sources might not include all the sources on the network; these will be added (or removed) as additional or new sources are discovered. It commonly takes a few seconds to discover all sources on a network._ https://docs.ndi.video/all/developing-with-ndi/sdk/ndi-find

Waiting is the proper thing to do, otherwise the user will waste time trying to track down connectivity issues including firewalls, when really it was just a matter of waiting. 

Also, timing tests have shown that the wait function only waits the full wait time when no new ones were found.  On a typical launch of the properties dialog, the wait times were 1ms, 6ms and then 1000ms for the last one. 

We could reduce the time down to 500ms, but they say a few seconds is needed. I could also add a refresh source list button and do the proper wait. In any event, I think a lot of problems people are having with the firewall is that we are not waiting long enough for the sources to show up.
Please let me know how to proceed with getting this change in place.

Trouffman:
I agree we should allow for some "extra time" whenever possible, but this cannot be at the expense of the UI loading.

Also referring to the documentation : 
"For applications that wish to list the current sources in a user interface menu, the recommended approach would be to create an NDIlib_find_instance_t instance when your user interface is opened and then – each time you wish to display the current list of available sources – call NDIlib_find_get_current_sources."

Which is why maintaining the list in "ndi_find" at the plugin level (and not at the source) would make sense.
> we can either have a "recurring" check for new source ( in background ), or this to be triggered for "extra check of new source" at different times (a function like "refresh source discovery" ?)

Delaying the UI display because we look for source will impact the user experience and make it look "sluggish". Which is why the "refresh source list" or "detect sources" button would be a preferred option overall.

Suggestion for now :
- implement ndi source finder at plugin level (load with plugin load, destroy at plugin unload)
- Have a function to "refresh the list" that can be called a different places of the plugin.
- Find a good place to call this refresh / wait on sources (even better if this is in the background) > for now we can keep it when launching the dialog box for lack of a better location.

Trouffman:
> Another option is to spin off a thread to do the waiting and update the properties list. This way the dialog is not delayed, but when or if the user has decided to select a source, the list will be complete. Adding a refresh button would merely spin off the same thread. The thread would end as soon as a complete list is received.

This would be a good option IMHO.

This way: 
- list of available source is updated at regular interval
- No delay in the UI

BitRate27:
I updated the PR to spin off a thread to wait for all the sources to be located. A new class was created called NDIFinder which has one public method called getNDISourceList. The method will return a list immediately if the list it has is less than 5 seconds old. Otherwise, a thread is created where it waits for all NDI sources to show up, waiting at most 1 second. When no new sources show up, a callback provided in the getNDISourceList call, is called with the list NDI names.

Trouffman (approved):
Tested locally on a MacBok Air. The behavior that needed to addressed seems to be fixed.
Also tested with test Pattern and a variety of source, did not had issue either.

Trouffman:
Tested on OBS31.0.3 on MacOS 15.4 / Windows 11 / Ubuntu 24.04 - no issues - assumed good.

Issue #1196, opened 2025-02-09
[Bug]: Changing profile with different canva/output resolution does not update the output/filter


Trouffman:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### Issue reported - Current Behavior

Changing a profile with a different output resolution does not update the output.

### Expected Behavior

All stream stop and are reloaded with the output resolution set on the profile.

### Steps To Reproduce

Tested on Windows 11 only for now.

Have 2 profile with each different "Output (scaled) resolution"

1. Open OBS
2. Enable NDI Output
3. Monitor the Feed (NDI Monitor or another OBS)
4. Change Profile
5. NDI feed do not stop / use the new resolution

### Environment

```markdown
- OS: Windows 11 24H2
- OBS: 31.0.1 
- DistroAV: 6.0.0
- NDI: 6.1.1.0
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [x] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

BitRate27:
I have confirmed we are not being notified when the profile changes. This explains why the NDI output is  @not changing when the profile is changed by the user. 

#Looks like we need to respond to this event: OBS_FRONTEND_EVENT_VIDEO_RESET which was added with this PR: https://github.com/obsproject/obs-studio/pull/11685.

We need to wait until this PR is pulled into OBS and we are building with it to close this issue. Since the change is adding a new event type, it will be backward compatible with prior versions, they just will not see the fix.

ddoerr:
I also came across this problem today. Tested on MacOS Sequoia. Disabling the NDI output, then changing the profile (1080 to 4k in my case) and reenabling the NDI output will also crash the NDI in the background. I have to restart OBS to get output again.

Trouffman:
This is now considered fixed with PR #1198 & associated bug fix in #1237

Issue #1195, opened 2025-02-09
[Bug]: Enabling NDI output after changing "Output (Scaled) Resolution" fails


Trouffman:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### Issue reported - Current Behavior

OBS crash when enabling output after changing Outpu (Scaled) Resolution in Settings.

No entries that much in the logs.

```
[DistroAV] +main_output_init()
[DistroAV] +main_output_start()
[DistroAV] main_output_start: starting NDI main output 'OBS'
[DistroAV] +ndi_output_start(name='OBS', groups='', ...)
```

Crash logs shows:
```
obs.dll!obs_output_begin_data_capture+0x3ff
distroav.dll!ndi_output_start+0x234
obs.dll!obs_output_actual_start+0x6b
obs.dll!obs_output_start+0xcf
distroav.dll!main_output_init+0x490
distroav.dll!OutputSettings::onFormAccepted+0x135
```

### Expected Behavior

Enable Output with the new resolution.

### Steps To Reproduce

Reproduce on Windows (did not reproduce in MacOS)

1. Launch OBS
2. Settings > Video > Output (Scaled) Resolution > change this
3. Click OK
4. Open "DistroAV NDI Settings"
5. Enable Main Output
6. Click OK

This should crash.

I suspect this is because we do not use the proper resolution (don't read properly and still reference the older resolution memory)

### Environment

```markdown
- OS: Windows 11 24H2
- OBS: 31.0.1
- DistroAV: 6.0.0
- NDI: 6.1.1.0
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [x] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

BitRate27:
I can duplicate this on Windows 11. It is indeed the same problem fixed in [PR 1148](https://github.com/DistroAV/DistroAV/pull/1148). There are two changes in that PR, the first one is to keep OBS from crashing, the other will not allow Main Output to be turned on if the format is not supported.
I can create a PR with just the first fix in it.

PR #1198, opened 2025-02-09
Recreate main output every time it is turned on

BitRate27:
The problem exists because an output would be created even when no main output was selected. When the user went to change the color format in the OBS Settings->Advanced dialog, or any setting in OBS Settings ->Video, this would invalidate the video output in DistroAV NDI Settings. Then when Main output was turned back on, OBS would crash because output->video is corrupted.

This makes the changes to main-output look big but it was mainly removing an indented if condition. In summary we always call main_output_deinit in main_output_init, and then if it is not enabled, we return, otherwise we create the output and start it every time it is enabled.

This fixes:
https://github.com/DistroAV/DistroAV/issues/1147
https://github.com/DistroAV/DistroAV/issues/1195


Trouffman:
When testing I get `Tried to reset when already active` in the log.

Also result in a black screen after changing the profile.

Can you check that the merge from master was correct?

BitRate27:
👍

BitRate27:
The `Tried to reset when already active` message was in the code before the PR. The resolution not changing when a new profile is chosen, while main output is active, is because we never responded to the OBS_FRONTEND_EVENT_PROFILE_CHANGING and OBS_FRONTEND_EVENT_PROFILE_CHANGED events. When these event types are added to the event callback for the plugin, and proper action is taken (calling main_output_deinit and main_output_init respectively), the main NDI output resolution changes, and as a bonus, we no longer get the `Tried to reset when already active` messages in the log. 
This should when pulled this should also resolve: https://github.com/DistroAV/DistroAV/issues/1196 and https://github.com/DistroAV/DistroAV/issues/1195.

src/main-output.cpp
```diff
@@ -124,44 +124,33 @@ void main_output_init()
 	auto output_groups = config->OutputGroups;
 	auto is_enabled = config->OutputEnabled;
 
-	if (output_name.isEmpty() || //
-	    (output_name != context.ndi_name || output_groups != context.ndi_groups)) {
-		main_output_deinit();
+	main_output_deinit();
 
-		if (!output_name.isEmpty()) {
+	if (is_enabled && !output_name.isEmpty()) {
```
Trouffman, line 129: This removes the "(output_name != context.ndi_name || output_groups != context.ndi_groups)" part of the if statement, do we remember why this was there?

paulpv: > do we remember why this was there?

Change detection of a user applies a settings change of the NDI output name or group.

No need to re-init the output when the settings are saved if the name/group does not change.

Maybe that check should be in `output-settings.cpp` `onFormAccepted()` and not build too much logic into `main_output_init()`; that would require it to save the name & group off to a temp and then check for a change and call output init as appropriate.
`output-settings.cpp` `onFormAccepted()` was cleaned up in 6.0.1-rc1:
https://github.com/DistroAV/DistroAV/commit/20e5c260eae8e7eb3f7d91881ea07a94635e7b3c#diff-450b5512730a34621354f8ff69c70be984d559191918e0553b1f89b7e4a21b75L235

src/plugin-main.cpp
```diff
@@ -307,6 +307,10 @@ bool obs_module_load(void)
 					// Unknown why putting this in obs_module_unload causes a crash when closing OBS
 					main_output_deinit();
 					preview_output_deinit();
+				} else if (event == OBS_FRONTEND_EVENT_PROFILE_CHANGING) {
```
paulpv, line 310: Is this related to the PR title?

BitRate27: Kind of...it is to fix two existing bugs found while testing this fix. We need to react to the profile changing event and close the output so we don't get the already initialized message (minor) and not recreating the main output with the new profile settings (major). We could create a new PR, but I instead referred to these (1195 and 1196) in the PR.

BitRate27: We need to leave the check for enabled in the main_output_init function so we can start the main output (if enabled) when we load from a config. onFormAccepted() is not the only way to initialize main output.

src/main-output.cpp
```diff
@@ -124,44 +124,33 @@ void main_output_init()
...
+		obs_log(LOG_INFO, "main_output_init: creating NDI main output '%s'", QT_TO_UTF8(output_name));
+		obs_data_t *output_settings = obs_data_create();
+		obs_data_set_string(output_settings, "ndi_name", QT_TO_UTF8(output_name));
+		obs_data_set_string(output_settings, "ndi_groups", QT_TO_UTF8(output_groups));
+
+		context.output = obs_output_create("ndi_output", "NDI Main Output", output_settings, nullptr);
+		obs_data_release(output_settings);
+		if (context.output) {
 			obs_log(LOG_INFO, "main_output_init: creating NDI main output '%s'", QT_TO_UTF8(output_name));
```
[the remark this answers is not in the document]
BitRate27: Fixed typo on line 138 (after check for output created) to say creat**ed** instead of creating.

src/main-output.cpp
BitRate27: Even though we need to leave the enabled check in the init, we can still check for changes in the output before initializing main or preview output. Added a commit to do this.

Trouffman:
Tested on Mac. Seems to fit the bill.
Changed between profiles (1080P / 4K) > the output is recreated as expected.

In case of issues it solves :
#1147 : this will show an error in the log file, avoid a crash but still have the "output enabled" ticked.
This could be improved with #1148 

#1195 : TBC

Trouffman (approved):

BitRate27:
Seems with this PR, we are no longer outputting Preview scenes. It preview shows up as a NDI source, but it doesn't output anything. Need to re-open and rethink this change. I believe it is related to Preview and Main using the same OBS output.

PR #1237, opened 2025-03-31
Fix the NDI Preview Output issue introduced in 1198

Trouffman:
This PR address the following :

- Fix the bug introduced in PR #1198 
- Added a log Entry to display the configuration submitted as a one-liner

How this fix the problem ?

- The Preview-output.cpp was not, modified to reflect the changes integrated in main-output.cpp

General hope is that the 2 codebase are extremely similar (with some extra code in Preview for specifics in there)

This bring the codebase to be similar for the approach & logic on both Main output and Preview output

### DIFF
--- a/src/forms/output-settings.cpp
+++ b/src/forms/output-settings.cpp
@@ -183,15 +183,23 @@ void OutputSettings::onFormAccepted()
 
 	config->AutoCheckForUpdates(ui->checkBoxAutoCheckForUpdates->isChecked());
 
+	// Output settings for debugging & diagnosis
+	obs_log(LOG_INFO,
+		"Output Settings set to MainEnabled='%d', MainName='%s', MainGroup='%s', PreviewEnabled='%d', PreviewName='%s', PreviewGroup='%s'",
+		config->OutputEnabled, config->OutputName.toUtf8().constData(), config->PreviewOutputGroups.toUtf8().constData(), config->PreviewOutputEnabled,
+		config->PreviewOutputName.toUtf8().constData(), config->PreviewOutputGroups.toUtf8().constData());
+
 	config->Save();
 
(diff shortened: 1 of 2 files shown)


BitRate27:
It looks like your changes are indeed needed as they are the exact changes I made to [PR 1148](https://github.com/DistroAV/DistroAV/pull/1148) a couple of days ago. However, I still do not get any preview output when I run this PR on the latest master branch from obs-studio. I can select the NDI preview output in Studio Monitor, but there is no video feed. When I change to another preview scene, the NDI preview output in Studio Monitor does not change. Please let me know if have trouble duplicating this issue.
0, Get into Studio Mode
1. In Tools->DistroAV NDI Settings, check Preview Output
2. Enter "OBS Preview" for the name
3. Select OK
4. Open NDI Studio Monitor
5. Select your local machine, and select "OBS Preview" (this being available is indicative of the NDI sender being created)
6. Error 1: the display does not change and does not match what is shown in the Preview window in OBS
7. In OBS select another scene
8. Error 2: the display does not change and does not match what is shown in the Preview window in OBS

BitRate27 (approved):
I tested this last change by: 
1. turning off main output, changing video resolution in settings, turning main back on.
2. turning off main output, changing color format in settings, , turning main back on.
3. changing profiles without turning off main output.
4. Used empty Main Output and Preview Output names.
No longer crashes on changes!

Trouffman:
Can you confirm that your initial reply in this thread was only when using the latest master branch of OBS ?

BitRate27:
The issue with preview output mentioned above was only seen with the master branch of obs-studio cloned on March 31, 2025. I did not see the issue and verified this PR by installing the .dll generated in the CI actions, on the 31.0.3 release of OBS Studio.

Issue #613, opened 2021-01-25
NDI source output filter does not recognize size changes from crop filter


TimPeterson:
Steps:
1. Create a scene with one source, for example an image.
2. Add a Crop/Pad filter to the image source, e.g. cropping from 1920x1080 down to 1280x720.
3. Add an NDI filter output to the image source.
4. View the NDI output in Studio Monitor.

Expected result:
Studio Monitor shows a 1280x720 signal with the cropped image, the same as shown within OBS.

Actual result:
Studio Monitor shows a 1920x1080 signal with the cropped 1280x720 image in the upper left corner of the frame.

Using obs-ndi 4.9.0 with OBS 26.1.1 on Windows 10 1909.

Trouffman:
Thanks for reporting and providing steps to reproduce.

Closing old issues.
Please use the latest version of DistroAV / OBS and open a new issue if this persist.

markusloffler:
I can reproduce this problem with OBS 31.0.3 on MacOS and DistroAV 6.0.0

BitRate27:
What are the exact numbers you put into the Crop/Pad filter?
Relative?
x?
y?
width?
height?

markusloffler:
Let's say I have a capture source of a window that has the size 1000 x 500. I set the crop filter to Left = 0, Top = 0, Right = 300, Bottom = 0. Then the NDI stream still has the size 1000 x 500, not 700 x 500. On the right side of the video stream, the pixels are just set to black with a width of 300.
It doesn't matter if the crop filter is before or after the NDI output.

BitRate27:
This is fixed in https://github.com/DistroAV/DistroAV/pull/1239 . It is a requirement that in order for the NDI output to be cropped, the Dedicated NDI output filter must be after/below the Crop/Pad filter in the list.

markusloffler:
Thanks a lot for the quick fix. Happy to help with testing.

This is what I get:
- Crop filter after NDI output: Behavior as before
- Crop filter before NDI output: Black output

BitRate27:
Does the size of the feed look correct? I saw the black screen with my own build of obs-studio but worked fine on a 31.0.3 install.

markusloffler:
Yes, size is correct (checked with NDI Video Monitor)

markusloffler:
FYI: For testing, I copied distroav.plugin from artifacts to ~/Application Support/obs-studio/plugins and enabled it with xattr -d com.apple.quarantine

[...]

Issue #969, opened 2024-01-28
[Bug]: 


Krisiries:
### Operating System Version

Windows 10 22H2

### CPU

Ryzen 7 5800X

### GPU

RTX 3080

### NDI Devices

OBS plugin only

### OBS Version

30.0.2

### OBS Installation Method

windows executable

### OBS-NDI Version

4.13

### OBS-NDI Installation Method

windows executable

### [Extra] Installation Steps

Source clone was used for testing, but is not required to achieve this effect.

### OBS Log [URL]

https://obsproject.com/logs/zJrZOCGriRB8EQdW

### NDI Version

5.6.0.0

### Describe the bug

When a scaling filter is applied before the NDI filter, NDI ignores the new size of the frame, and transmits the original size.

For example: if a 2560x1440 source has a scaling filter to 1280x720, then an NDI filter is applied, the client computer receives a frame that is 75% transparent with the actual content located in the top left, regardless of position of the source.

Host computer with main screen and scaled clone of main screen shown:
![image](https://github.com/obs-ndi/obs-ndi/assets/640422/c29fb40c-1734-4af8-8815-d9ec80ac1640)
Client computer with 720p preview and empty received frame shown:
![image](https://github.com/obs-ndi/obs-ndi/assets/640422/35c07ca9-4b4e-4579-9b8f-241c07c0ab9f)

### Steps to reproduce the problem

1. Create a source that is either a source clone of current scene or copy of another scene.
2. Apply a "Scaling/Aspect Ratio" filter to the new source
3. Apply the "NDI Output" filter to the same source, below it in the list
4. Attach a client pc to the "NDI Output" filter
5. Frame received is the original size of the source created in step 1 

### Expected behavior

Frame size should be consistent with the "Scaling/Aspect Ratio" sizing if the "NDI Output" filter falls after the scaling filter.

Krisiries:
Tested with a build that changes lines 166 and 167 in _obs-ndi-filter.cpp_. 

To get the post-filter resolution, width and height instead of base width and height would need to be grabbed, however both of these solutions are making an assumption that the NDI filter is first (in the current implementation) or the filter is last (in my test):

If NDI Output filter is before scaling (current implemenation):
	uint32_t width = obs_source_get_base_width(target);
	uint32_t height = obs_source_get_base_height(target);

If NDI Output is after scaling (test fix for my local use):
	uint32_t width = obs_source_get_width(target);
	uint32_t height = obs_source_get_height(target);

[...]

PR #1239, opened 2025-04-04
Use previous filter as source for dedicated NDI output filter

BitRate27:
This PR fixes a problem reported in https://github.com/DistroAV/DistroAV/issues/613. The NDI output filter was not using the cropped image output from the Crop/Pad filter before it in the filter list. The fix was as simple as getting the target source (previous filter output) from OBS, instead of the parent source (the actual source).

Verified the size of the NDI output from OBS changes as the size of the crop changes in OBS. Also removed the Crop/Pad filter and verified the full source size was output over NDI.


Trouffman (dismissed):

Trouffman:
This is pretty crazy the fix was that simple!

Reading the doc for this totally makes sense.

Trouffman:
@BitRate27 can you check if this would have an impact/solution on this : https://github.com/DistroAV/DistroAV/issues/969

BitRate27:
The user who reported the problem, tried out this fix and sees a black NDI feed output. I can also duplicate this problem. After further investigation, it seems our use of the medio-io/video non-public API makes this solution unusable.

Trouffman:
@BitRate27 maybe this would be a good start to steer away from the non-public/undocumented media-io/video API.

You can get in it in this PR or create a new one and close this one.

Trouffman:
THis has been managed in #1343

PR #1251, opened 2025-04-23
Update Plugin dependencies to OBS-Deps 2025-03-06

Trouffman:
This PR is "ahead" of the obs-plugin-template but relies only on tagged release by master OBS Project.

- Prebuilt & QT6 pre-built : 2025-03-06

This is aimed to stay as close as possible to OBS Studio and be ready (or contribute to the obs-plugintemplate)

This is not intended to be merged until fully tested to not move away too much from the obs-plugintemplate.


Trouffman:
Too many changes in the same PR would create confusion. changes will be split in their respective PR.

Trouffman:
This PR now focus only on the new deps update.

Note : this will require to squash all commit messages as well

Trouffman:
This PR is deprecated in favor of #1318

PR #1256, opened 2025-04-25
Enhance Log Entry and add unique code for ERROR & WARNING level

Trouffman:
Additional reformatting of the Log entries for consistency & further add unique identifier for each unique ERROR/WARNING level.

## Goal
- Easier to search solution for a specific error
- Less confusion between error log entries
- Could help to automate basic support / log analysis
- Numerical value is irrelevant to other similar 3digit error code (like HTTP)

## What has been done
- Further enhance previous Log Entry reformatting
- All ERROR level message now have a unique identifier : "ERR-XXX"
- All WARNING level messages now have a unique identifier : WARN-XXX

## Format & consistency rules
- WARN-XXX for LOG_WARNING level entries
- ERR-XXX for any LOG_ERROR level entries
- XXX is 3 digit
- XXX does not relate to any other error code that exists (like HTTP Code)
- Some XXX ID can be used of left for special (fun) purpose
> like 404 when the NDI library is not found) but is discretionary
- XXX value is unique across any LOG_LEVEL
- If a downgrade from ERROR to WARNING : the 3 digits stay the same
- If an upgrade from WARNING to ERROR : the 3 digits stay the same
- If a downgrade to a non-WARNING/ERROR level, keep the log Entry ID. (INF-)
> In some cases this can be move to the end of the log-entry
> If not-upgraded within 12months AND more than 1 Major releases AND in need for cleanup : Code can be "retired" but must still be accessible for historical purposes (wiki?).
- If the Message attached change : the ERROR code stay the same
- Log Entry are Specific to the error / location in code / Trigger reason / Root cause. Therefore, If the code change, evaluate if this is the same error / log entry or a new one.
- In case we go through more than 999 Error/warning ID... The project will be VERY successful, just add a digit.

## To-Do
- Add ERROR Code to the WIki for the relevant Issues / troubleshooting steps
- Only use the code/ID, not the LOG-ENTRY message (easier to maintain)

### ERROR Code ID
ERR-400 - Failed to start NDI Main Output
ERR-411 - Failed to initialize NDI Main Output
ERR-412 - Failed to create NDI Main Output

ERR-410 - NDI Output cannot start : Unsupported pixel format

ERR-407 - Error creating the NDI Receiver
ERR-408 - Error creating the NDI Frame Sync
ERR-430 - NDI Source uses an unsupported video pixel format

ERR-404 - NDI library not found, DistroAV cannot continue
ERR-401 - NDI library failed to load
ERR-403 - OBS-NDI is detected and needs to be uninstalled before DistroAV can work
ERR-406 - NDI library could not initialize due to unsupported CPU
ERR-405 - Error loading the NDI Library from path
ERR-402 - Error loading QLibrary with error

ERR-409 - Check for update failed: Cannot get the main OBS window

### Warning Code ID
WARN-413 - NDI Output could not start. No Audio/Video data available.
WARN-415 - NDI Sender data capture failed
WARN-416 - NDI Sender initialisation failed

WARN-414 - Invalid or unknown behavior detected in config file
WARN-423 - Failed to create NDI Preview Output 

Update Error Code that might need rework.
WARN-417 - Update Check received a non-expected response
WARN-418 - Update Check took too long for a response
WARN-419 - Update Check request failed
WARN-421 - Update Check could not open the update file
WARN-422 - Update Check could not verify the update file

### DIFF
--- a/src/forms/update.cpp
+++ b/src/forms/update.cpp
@@ -208,15 +208,15 @@ QString GetObsCurrentModuleSHA256()
 	auto module = obs_current_module();
 	auto module_binary_path = obs_get_module_binary_path(module);
 #if 0
-	obs_log(LOG_INFO,
-	     "GetObsCurrentModuleSHA256: module_binary_path=`%s`",
+	obs_log(LOG_DEBUG,
+	     "GetObsCurrentModuleSHA256: module_binary_path='%s'",
 	     module_binary_path);
 #endif
 	QString module_hash_sha256;
 	auto success = CalculateFileHash(module_binary_path, module_hash_sha256);
 #if 0
-	obs_log(LOG_INFO,
-	     "GetObsCurrentModuleSHA256: module_hash_sha256=`%s`",
+	obs_log(LOG_DEBUG,
+	     "GetObsCurrentModuleSHA256: module_hash_sha256='%s'",
 	     QT_TO_UTF8(module_hash_sha256));
 #endif
 	return success ? module_hash_sha256 : "";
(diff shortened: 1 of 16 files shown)


src/forms/update.cpp
```diff
@@ -521,7 +524,9 @@ bool updateCheckStart(UserRequestCallback userRequestCallback)
 	auto timer = new QTimer();
 	timer->setSingleShot(true);
 	QObject::connect(timer, &QTimer::timeout, []() {
-		obs_log(LOG_WARNING, "updateCheckStart: timer: Request timed out");
+		obs_log(LOG_WARNING, "WARN-418 - Update Check took too long for a response");
+		PostToMainThread("timer->timeout", []() { updateCheckStop(); });
```
Trouffman, line 528: I need to check where this is coming from.

Trouffman: This is coming from an old commit : https://github.com/DistroAV/DistroAV/commit/f3d6229fcb3b57dc22a326342eea4b6f5838b301

but flagged as new due to the changes in the lines counts / order

PR #1261, opened 2025-05-04
Add proper UYVA support

Lordmau5:
Currently UYVA is being treated as UYVY in DistroAV.

This makes sense as OBS does not currently support NDI's special UYVA format and the way UYVA is handled is in 2 planes ([Reference](https://docs.ndi.video/all/developing-with-ndi/sdk/frame-types)):
- The first plane is effectively just UYVY, meaning that images will still retain their color, **however**
- The second plane has the alpha information

Without support for said second plane the image will just look like it's on a black background:
image

However, adding support in both OBS through [this PR](https://github.com/obsproject/obs-studio/pull/12125) and also within DistroAV will make it look correctly:
image

This PR will not build without the OBS PR merging first as it tries to access a new video format.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -742,8 +742,11 @@ void ndi_source_thread_process_video2(ndi_source_config_t *config, NDIlib_video_
 		break;
 
 	case NDIlib_FourCC_type_UYVY:
-	case NDIlib_FourCC_type_UYVA:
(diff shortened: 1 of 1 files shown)


Trouffman:
This PR Is dependent on OBS feature PR : https://github.com/obsproject/obs-studio/pull/12125

PR #1263, opened 2025-05-07
Add check for minimum NDI & OBS version requirement - 6.1 release

Trouffman:
# Context
Prepare for 6.1 release
Add Minimum version check (OBS & NDI)

# Goal
Enforce minimum requirements in OBS Major version and NDI SDK
Unload plugin if minimum requirements are not met (avoid issues & breaking things)

# Why enforce min requirement for OBS/NDI version?
Read more at : https://github.com/DistroAV/DistroAV/discussions/1189

## TL;DR :

> Updated Minimum Requirement
> 
>     OBS v31+
>     NDI SDK/Runtime v6+
> 
> Why is this change necessary ?
> There are changes integrated in OBS 31 that split the configuration files and we are following the guidelines set by the OBS team for a sustainable OBS-user community.
> 
> NDI v6 has been widely available for months now and proven reliable. The codebase in DistroAV 6 already integrate with NDI v6, even tho we did not leverage new feature and code optimization (yet), this will be a possiblity at "anytime" with DistroAV 6.1.

Adding a note to this TL;DR above : some changes in the audio processing in the DistroAV code requires it and bugs relating to audio have been corrected in NDI 6.1 (still we are only enforcing 6.0 for now).

# Behavior expected
If the minimum OBS & NDI version are not met:
- The plugin will unload
- An error entry in the log will be displayed with a unique error code that can be easily searched for
- TODO : Add a more comprehensive solution in the wiki for the error code 424 & 425 (minimum version not met)
- TBC : An error popup will inform the user with the error code

If the minimum OBS & NDI versions are met:
- Happy days! It will load as expected.

# Change management / transparency / Roll-out plan
- Some users are still on 4.14, and have not yet moved to DistroAV v6, but have moved on to OBS 31 and NDI 6. So no more excuses to hold back upgrade. The 4.14 codebase is over a year old and 2 major versions behind and not maintained/supported.

- Some distros (Fedora?) are quite behind on NDI SDK (like 5.6), I have reached out to a couple maintainers of said package/distros (Nobara) that this might impact some of their users. - Fedora is not officially supported, so this was as far as I could go.

- OBS 31 config management changes means that once the plugin is used on that version onwards, there is no way back for the config. No backward compatibility / downgrading will be put in place (as more than 2/3 of the users are on OBS 31+ already)

- There are a lot of bug fixes and improvement in 6.1 that would have deserved a minor release. But here we are, the longer you wait to update, the more changes are required.

- It is expected that upgrade from DistroAV 6.0 to 6.1 to be seamless. It is expected the same from 4.14 to 6.1 (but not planned to be tested)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ DistroAV (Formerly OBS-NDI)
 - **NDI Filter** (a.k.a. **NDI Dedicated Output**) : Transmit a single OBS source or scene audio to NDI
 
 ## Requirements
-* OBS >= 30.0.0 (Qt6, x64/ARM64/AppleSilicon)
+* OBS >= 31.0.0 (Qt6, x64/ARM64/AppleSilicon)
 * [NDI Runtime >= 6](https://github.com/DistroAV/DistroAV/wiki/1.-Installation#required---ndi-runtime)  
 * [Remove old OBS-NDI plugin](https://github.com/DistroAV/DistroAV/wiki/OBS%E2%80%90NDI-Is-Now-DistroAV)
 
--- a/buildspec.json
+++ b/buildspec.json
@@ -38,7 +38,7 @@
     },
     "name": "distroav",
     "displayName": "DistroAV",
-    "version": "6.0.0",
+    "version": "6.1.0",
     "website": "https://distroav.org",
     "author": "DistroAV",
     "email": "contact@distroav.org",
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -260,10 +260,9 @@ bool obs_module_load(void)
 			obs_get_version_string(), PLUGIN_MIN_OBS_VERSION);
 
 		auto title = "OBS version not supported";
-		auto message = "Error-424: Plugin requires OBS " +
-			       QTStr(PLUGIN_MIN_OBS_VERSION) + " or higher <br>";
+		auto message = "Error-424: Plugin requires OBS " + QTStr(PLUGIN_MIN_OBS_VERSION) + " or higher <br>";
 		showCriticalUnloadingMessageBoxDelayed(title, message);
-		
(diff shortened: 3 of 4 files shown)


BitRate27:
Can the minimum version be dependent on the platform? Since the Fedora platform is behind, do we wait for that platform to move up before enforcing the minimum for all, or do we say Fedora will work with the minimum supported on that platform (5.6)?

Trouffman:
Sadly that minimum requirement is a hard requirement as it will impact on the configuration files, some functions and even some fixes

This makes DISTROAV 6.1 a sane base to built upon.

Also NDI 6 was technically already required for distroAV 6.0 but was not enforced.

For the Fedora users : I contacted the maintainer of the libndi package, he will look into it in the coming weeks.


This brings a broader question/issue : should we maintain the packages (libndi / DISTROAV) for the distro we officially support? Fedora is technically not officially supported, but we try to get it working still.

In short : no platform-specific exception should be implemented.

Should we have a way to allow for pre NDI V6 ? Nop, It was already required for distroAV 6.0. now enforced. Otherwise we are stuck to this feature level until the next DISTROAV major release 

Should we have a way to allow for pre OBS 31 ?
Plain and simple : not possible :D the plugin deps are 31.0.0.

Theses changes makes the reason for a new DistroAV major version.

Also : anyone on 4.14, this might be the last version where the "upgrade" will be supported.

RayneYoruka:
After the quick update:
Working fine on Ubuntu 24.04
Working as well on Fedora 42.. the locations for Fedora with the Ubuntu artifact:

`/usr/share/obs/obs-plugins/distroav/` < Locale.
`/lib64/obs-plugins/ `< Plugin location.

BitRate27 (approved):

PR #1232, opened 2025-03-26
Performance improvement on NDI Filter

Marsic1:
fixed encoder overload issue reported on https://github.com/DistroAV/DistroAV/issues/1231

PR #1266, opened 2025-05-07
NDI Output filter - Performance tweaks

Trouffman:
Refactor video frame handling in ndi_filter_offscreen_render to be closer to Preview Output

# Goal
Update the NDI filer codebase to be updated and more similar to the NDI Preview Output rendering.

# Context
* This was triggered by a discussion posted by @theotherraycole at : https://github.com/DistroAV/DistroAV/discussions/872
* Looked at the code and the only other place where something similar was used was for the NDI Preview output.
* Then re-used most of the code from Preview output to this.

Disclosure : I do not understand it all for the logic (i am not skilled in anything using gs / drawing textures, etc).

This will stay as a draft until more pairs of eyes can review it.2

### DIFF
--- a/src/ndi-filter.cpp
+++ b/src/ndi-filter.cpp
@@ -184,23 +184,22 @@ void ndi_filter_offscreen_render(void *data, uint32_t, uint32_t)
 			f->known_height = height;
 		}
 
-		video_frame output_frame;
+		struct video_frame output_frame;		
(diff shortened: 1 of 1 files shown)


Trouffman:
This PR is very similar to #1232

Trouffman:
This is superseeded by #1279

PR #1268, opened 2025-05-09
 Split configuration files between User & Global Store - OBS 31 requirement

Trouffman:
Built on the great work by @kaechele :

> This properly splits settings for this plugin into the new split configuration stores for app and user settings.
> 
> All user-editable settings go into the user store, all functional, non-user facing settings go in app store (e.g. timestamp of last update check).
> 
> On OBS versions prior to 31 there is no change in behaviour and all settings continue to be read from and stored to the global store only.
> 
> Fixes https://github.com/DistroAV/DistroAV/issues/1121

This was originally dismissed as the backward support was under discussion. With DistroAV 6.1: OBS 31 is required. So that backward support is no longer required.

As to grant @kaechele the praise for this great piece (especially the migration settings) this is a revival of the PR #1152 that was originally closed.

Things to address:
- Remove complexity of backward compatibility with pre-OBS 31
- Update / Add a few missing config entries
```
PARAM_SKIP_UPDATE_VERSION
Can be set & influenced by the user, the decision is to make it a "User" config

The other update elements are not influenced / presented to the User and will be in the App Config:
PARAM_LAST_UPDATE_CHECK
PARAM_MIN_AUTO_UPDATE_CHECK_INTERVAL_SECONDS
PARAM_AUTO_CHECK_FOR_UPDATES
```

Note : this PR was re-created with a branch that can be modified.


src/config.cpp
```diff
@@ -152,12 +196,14 @@ Config::Config()
 	  TallyPreviewEnabled(true)
 {
 	ProcessCommandLine();
-	SetDefaultsToGlobalStore();
+	SetDefaultsToUserStore();
+	if (obs_get_version() >= MAKE_SEMANTIC_VERSION(31, 0, 0))
```
Trouffman, line 200: Maybe add a check/config entry when that migration is done so it does not trigger at every plugin launch

BitRate27 (approved):

Issue #1285, opened 2025-06-06
"DistroAV" is not displayed in the OBS toolbar. (Spanish + Mac bug)


RayneYoruka:
### Discussed in https://github.com/DistroAV/DistroAV/discussions/1284

<div type='discussions-op-text'>

<sup>Originally posted by **DobleT04** June  5, 2025</sup>
Hello, **I'm a Spanish-speaking user**.
I installed DistroAV, but it doesn't appear in the top toolbar.
I can add NDI sources, but I can't send an NDI signal from my program.

Installation order:
obs-studio-31.0.3-macos-apple
NDIToolsInstaller6.2
distroav-6.1.0-macos-universal

Note: This is a clean install of OBS after formatting macOS 15.5 (24F74) - Apple M1 Pro

Thank you for your help.

![Error](https://github.com/user-attachments/assets/7fee5c0c-b066-4afc-90df-74a99024baad)
</div>

> After searching through all the modules, I managed to find the DistroAV NDI settings. It's now in Preferences.

![lol](https://github.com/user-attachments/assets/365e37e9-a89f-4845-9ed3-19f9e85d3491)

This is a very strange issue, as mentioned by @BitRate27 : 

> I can duplicate this by setting my language preference to Espanol. When the language is Espanol, selecting Preferences.. takes you directly to the DistroAV NDI Settings dialog and you can no longer access the OBS Preferences dialog. More tragic is that you have no way to set the language back to English at this point. There is no entry for DistroAV NDI Settings on the Tools menu either.

> Some other observations, works fine on Windows when you select Espanol. What is weird is that on Mac, the OBS menu has the menu button labeled "Preferences..." (in English) which brings up our settings dialog (the bug). It should be labeled "Adjustes" when Espanol is selected.

Created the bug report to keep track!

BitRate27:
Seems to only be a problem with es-ES and ca-ES. Other translations are fine.

RayneYoruka:
I speak es-ES. Perhaps I can help with the translation?

BitRate27:
Problem also exists in DistroAV 6.0.0, so doesn't seem to be a related to the new 6.1.0 release. If DistroAV is not installed the Preferences... button does bring up the main OBS Preferences dialog. 

[...]

PR #1288, opened 2025-06-07
Don't hijack OBS preferences dialog on MacOS in Espanol or Catalan

BitRate27:
Change Configuració and Configuración to Ajustaments and Ajustes in the ca-ES and es-ES locale files. 

When using the Configuració(n) keyword as the first word in the menu button text, MacOS (or some UI level library) thinks that menu button is the "Preferences" dialog for the whole application (OBS). Using Ajustes and Ajustaments doesn't cause the hijacking.

This fixes this bug: https://github.com/DistroAV/DistroAV/issues/1285


Trouffman (approved):
Seems straight forward.

Issue #1290, opened 2025-06-08
[Bug]: Installing 6.1.0 on Windows does not uninstall 6.0.0


BitRate27:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### OBS Log [URL] - Important!

https://obsproject.com/logs/K0kWPYKfhwom7xFF

### Issue reported - Current Behavior

If 6.1.0 version of DistroAV is installed with the Windows installer and the previous version of DistroAV (6.0.0) is already installed, there will be two DistroAV plugins installed.

### Expected Behavior

The old 6.0.0 version of DistroAV should be uninstalled when installing 6.1.0.

### Steps To Reproduce

1. Install DistroAV 6.0.0 using Windows installer
2. Install DistroAV 6.1.0 using Windows installer
3. Open OBS

![Image](https://github.com/user-attachments/assets/41aa4375-6132-4d06-a0d9-8aa079b6ce48)

### Environment

```markdown
- OS: Windows 11
- OBS: 31.0.3
- DistroAV: 6.1.0
- NDI: 6.1.1.0
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [ ] I have tested using a new scene collection to reproduce the issue
- [ ] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

tanelaavistu:
Same here.

Trouffman:
Thanks for reporting!

We are already looking to a fix for this.

The root cause has been identified as well : the new install location and installer does not "remove" the old one.

Please subscribe to this issue to be informed when fixed.

Trouffman:
@tanelaavistu This is now fixed in master and will be released very soon. Stay tuned !

tanelaavistu:
> @tanelaavistu This is now fixed in master and will be released very soon. Stay tuned !

Thank you!!

tanelaavistu:
@Trouffman thank you! You guys are fast.

Trouffman:
This has been fixed & released : get it at https://github.com/DistroAV/DistroAV/releases/latest

PR #1293, opened 2025-06-11
Uninstall DistroAV 6.0.0 when installing 6.1.0+

BitRate27:
Uninstall and directly remove DistroAV 6.0.0 files from the old Program Files location.

Starting with DistroAV 6.1.0, we now install DistroAV in OBS's ProgramData folder. This means the old install in Program Files is not removed/updated to the new version. This results in the plugin showing up twice in the Tools pulldown, and also unclear which is used for NDI sources.

To fix this we uninstall DistroAV 6.0.0 by finding the uninstall exe and running it. We also remove any files that may have been installed manually from the OBS's Program Files folder.

This is not a selectable task like the obs-ndi plugin removal.

This fixes issue https://github.com/DistroAV/DistroAV/issues/1290

### DIFF
--- a/cmake/windows/resources/installer-Windows.iss.in
+++ b/cmake/windows/resources/installer-Windows.iss.in
@@ -36,7 +36,7 @@ Name: RemoveOBSNDI; Description: "Remove OBS-NDI plugin"; GroupDescription: "Cle
 Name: AddFirewallEntry; Description: "Allow OBS/DistroAV in Windows Firewall (requires admin privileges)"; GroupDescription: "Post-Install Actions";
 
 [Files]
-Source: "..\release\Package\{#MyProjectName}\*"; DestDir: "{app}"; Flags: recursesubdirs createallsubdirs
+Source: "..\release\Package\{#MyProjectName}\*"; DestDir: "{app}"; Flags: recursesubdirs createallsubdirs ignoreversion
 ; NOTE: Do not use "Flags: ignoreversion" on any shared system files
 
 [Icons]


Trouffman:
This will not work as it is today.

Reason :
- The uninstall is called at the **end** of the installation of DistroAV. The uninstall is targeting the AppID, BUT 6.0.0 and 6.1.0 share the same ID.
That means : The installer will install 6.1.0, then delete 6.1.0. leaving 6.0.0 intact & removing 6.1.0.

We should uninstall 6.0.0 as a "pre-install" step (via uninstaller).
However, it is fine to keep the "files clean-up" (removing the files directly) as a "post-install" step.

While digging into this I found out that DistroAV 6.1.0 doe snot set the proper registry key (which is why this might have worked for you without issue).

We now have 2 issues to fix :
- 6.0.0 is not uninstalled
- 6.1.0 is not installed properly on Windows Registry (Key is {}_is1)

Trouffman:
Tested as follow:
- DistroAV 6.0.0 installed
- DistroAV 6.1.0 installed

Run This PR Installer with /LOG flag.

Detection of 6.0.0 & 6.1.0 works, Uninstall return code 0 (success)
```
2025-06-11 18:08:36.565   DistroAV plugin version already installed detected : 6.0.0...
2025-06-11 18:08:36.565   Uninstalling previous DistroAV plugin version...
2025-06-11 18:08:36.767   Uninstall previous DistroAV return code: 0
2025-06-11 18:08:36.767   Uninstalling previous DistroAV plugin version 6.1.0...
2025-06-11 18:08:36.947   Uninstall previous DistroAV return code: 0
2025-06-11 18:08:36.947   Starting the installation process.
```

Confirmed that 
* 6.1.0 - Registry entry is gone
* 6.1.0 files are removed as well

Issue :
* 6.0.0 - Registry entry is gone (not supposed to)

* A new entry in registry is "FF773A8D-7648-4630-8155-3D5190CDF785" - Expected to be "{FF773A8D-7648-4630-8155-3D5190CDF785}".

Trouffman:
Good outcome, it now look cleaner in the Uninstall list : 
image

Trouffman:
The latest commit before this message fix the {{APPID}} issue.

But brings to light that the Setup WILL use the previous Directory.

This can be fixed with : `UsePreviousAppDir=no` (default is yes) and this will not try to use the previous install location.

Trouffman (approved):

Issue #1015, opened 2024-05-09
[Tweak] Update the Linux libNDI helper script to install by default


Trouffman:
To simplify the installation process /debug on Linux we could improve the libndi-get.sh script to be "install" by default.

If this script must be used by other scripts, the "install" argument could be a "noinstall" instead.

Trouffman:
This will not be implemented because this is communicated properly in the documentation.

PR #1304, opened 2025-06-17
Update Linux Installer Script

Trouffman:
Initially to address address #1015 but as @paulpv mentionned a couple time, it is not the best practice to have the script auto-install by default.

- Update Linux Installer Script with doc on how to use
- Update package creator to use NDI 6.2

Things to discuss & check before commit:

- [x] No impact detected in the Github Actions
- [x] No impact detected for building the package (not supposed to)
- [x] No impact detected in third-party packager (mostly Flatpak here)
- [x] Confirm we (the project) are OK with this being a "noinstall" by default option
- [x] Update relevant Wiki pages

### DIFF
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -420,7 +420,9 @@ const NDIlib_v5 *load_ndilib()
 	// Flatpak install will look for the NDI lib in /app/plugins/DistroAV/extra/lib
 	locations << "/usr/lib";
 	locations << "/usr/local/lib";
+#if defined(Q_OS_LINUX)
 	locations << "/app/plugins/DistroAV/extra/lib";
+#endif
 #endif
 	auto lib_path = QString();
 #if defined(Q_OS_LINUX)


Trouffman:
This will also look for the libndi files in an additional location to help with the Flakpak release at https://github.com/flathub/com.obsproject.Studio.Plugin.DistroAV

Once released the following code can be removed from the Flatpak process:
https://github.com/flathub/com.obsproject.Studio.Plugin.DistroAV/blob/fc770add1c1a0ded47f34cbc199dd46892c4c99a/com.obsproject.Studio.Plugin.DistroAV.json#L90-L93

Trouffman:
For the record and the changes to add the flatpak path to the detection logic:

From a Discord discussion with tytan652 (flatpak maintainer):

> In the context of a Flatpak application, we could make ld to check additional paths so QLibrary("ndi") could work.
> In an extension we can't with extra-data [...]. The full path is the only way for the plugin Flatpak

RayneYoruka:
Have you tested it already? Asking to approve.

Trouffman:
Install Script Tested on Ubuntu 24.04.
Built Package tested and validated as well on Ubuntu 24.04.

PR #1314, opened 2025-06-21
Fix plugin duplicate detection

paulpv:
## Summary
- detect duplicate installs of DistroAV using `count_modules`
- log and show error dialog when another copy is present
- add translations for duplicate plugin error

## Testing
- `git status --short`

------
https://chatgpt.com/codex/tasks/task_e_68571621dffc83338128eea87ab94f4b


Trouffman:
Could this be simplified ?

As on :
- no need to count the number of instances detected, as long as there is "another one" the error should trigger.

Aka :
- plugin can be first to load OR not
- plugin need to detect:
1. if another (same) plugin is already loaded (like obs-ndi)
2. if another (same) plugin is trying to load.

Trouffman:
Seems this will be managed within OBS at some point : https://github.com/obsproject/obs-studio/pull/12285

Trouffman:
Until this is a fucntionality in OBS, maybe we can leverage : obs_enum_modules and obs_find_module_file.

Here is a revised approach :

### Goal : 
As part of early check / loading of the plugin :
- Detect if another version of DistroAV has been loaded already by OBS.
- Return the path to the conflicting (obs-ndi or distroav) module already loaded
- When DistroAV detected : Inform user that the plugin should be installed only once (redirect to the Wiki for support)
- When obs-ndi detected : Inform user to remove OBS-0NDI (redirect to wiki for support)

### Approach
- Use the same type of check for the current plugin (with obs_module_name()) and other conflicting plugins (obs-ndi in our case) - this could leverage `is_module_found` or a `obs_enum_modules `
- For each conflicting checks : return the installation path / file name of the modules (can leverage `obs_get_module_lib` or `obs_module_file` or `obs_get_module_file_name` or `obs_get_module_binary_path`

Trouffman (approved):
Let's confirm if this is still required with OBS 32 around the corner!

Trouffman:
This might not be required with the upcoming OBS 32 plugin manager.

PR #1315, opened 2025-06-28
Revert buildspec to OBS upstream.

Trouffman:
This should be taking care of reverting #1252 back to the proper upstream.

If this is successful, it should also be updated on the DistroAV's fork of OBS-Studio for the sake of building the plugin.


Trouffman:
And it suceeded !

Issue #1316, opened 2025-07-04
[SR]: Automatically Deactivate Scene When NDI Signal Is Lost


Xzensi:
### This project is community-driven and does NOT provide support.

- [x] I understand that DistroAV does not provide support and this request might be closed for any reason without resolution.

### Type of Support needed

Configuration

### Before you submit a support request, confirm the following:

- [x] I have checked this is not an existing active bug/feature request
- [x] I have checked that my OS / OBS / NDI versions are supported
- [x] I have followed the NDI Runtime installation steps as described in the Wiki
- [x] I have followed the DistroAV installation steps as described in the Wiki
- [x] I have tested without firewall
- [x] My version of DistroAV is up-to-date
- [x] My version of OBS is up-to-date
- [x] My version of NDI Runtime is up-to date
- [x] OBS launch without error message (User Interface/popup)
- [x] OBS log file does not show any errors
- [x] I can access the DistroAV settings menu
- [x] The “NDI Source” option exist & a new 'NDI Source' can be added
- [x] OBS / NDI / DistroAV are allowed to use the network connection (Firewall)
- [x] The NDI feed can be viewed using 'NDI Studio Monitor'
- [x] The NDI feed can be received in OBS
- [x] A working NDI feed can be received in 'NDI Studio Monitor'
- [x] A working NDI feed can be received in OBS

### OBS Log [URL]

.

### I need help with...

I'm wondering if there's currently any way — either natively or via scripting — to **automatically deactivate or hide a scene** (or a source within a scene) **when an NDI source loses signal** or goes offline?

My goal is to have a fallback scene (e.g., "Technical Difficulties") show up **only when the NDI source is no longer sending video or becomes unavailable.** I’ve tried using Lua scripting to monitor the resolution of the source (checking for 0x0), but it doesn’t seem reliable, possibly due to how the NDI source is cached or frozen on the last frame.

### Environment

```markdown
- OS: Windows
- OBS: Latest
- DistroAV: Latest
- NDI: Latest
```

paulpv:
@Xzensi That's a pretty good idea. I wonder if there is also a way to extend Advanced Scene Switcher (https://github.com/WarmUpTill/SceneSwitcher) macros to first class recognize NDI events and perform set actions.

paulpv:
As an idea for a code solution: 
https://chatgpt.com/s/cd_686894eda3788191923e414ddfae0f24

Ignore the readme and whitespace changes. 
[...]

PR #1317, opened 2025-07-05
Implement automatic source disable/enable on NDI disconnect/connect

paulpv:
Per #1316
> I'm wondering if there's currently any way — either natively or via scripting — to automatically deactivate or hide a scene (or a source within a scene) when an NDI source loses signal or goes offline?
> 
> My goal is to have a fallback scene (e.g., "Technical Difficulties") show up only when the NDI source is no longer sending video or becomes unavailable. I’ve tried using Lua scripting to monitor the resolution of the source (checking for 0x0), but it doesn’t seem reliable, possibly due to how the NDI source is cached or frozen on the last frame.

In short, they want to automate some things (scene/source) based on if the NDI source is available or not.

This PR disables the DistroAV/NDI source if the NDI stream is lost, and enables the DistroAV/NDI source if the NDI stream is obtained.

Disabling/Enabling the source can also help with further automation via OBS Advanced Scene Switcher:  
https://github.com/WarmUpTill/SceneSwitcher

## Summary
- add translated string for auto deactivation option
- expose `ndi_auto_deactivate` property and default value
- control source enable state using a `source_enabled` flag so calls to `obs_source_set_enabled` only happen when needed

------
https://chatgpt.com/codex/tasks/task_e_686893a590008333bdd2b29f81f27263


paulpv:
This might fix https://github.com/DistroAV/DistroAV/issues/1316

This PR is [after much human prodding] AI generated **AND COMPLETELY UNTESTED!!**

src/ndi-source.cpp
```diff
@@ -546,15 +554,29 @@ void *ndi_source_thread(void *data)
 		// check if there are any connections.
 		// If not then micro-pause and restart the loop.
 		//
-		if (ndiLib->recv_get_no_connections(ndi_receiver) == 0) {
+		bool no_connections = ndiLib->recv_get_no_connections(ndi_receiver) == 0;
+		bool current_enabled = obs_source_enabled(s->obs_source);
```
paulpv, line 558: This seems undesirable for every loop to test for source enable/disable. :/

Trouffman:
@paulpv can you edit the OP to detail the pain-point it is addressing, and the approach taken.

paulpv:
> @paulpv can you edit the OP to detail the pain-point it is addressing, and the approach taken.

Done

Trouffman:
Just tested this :
It does clean-up the frame, BUT this does NOT "disable" the source.

This behaviour can be good too, but we need to rename the feature.

Can be renamed :"Remove last received frame when NDI signal is lost" (or a better option)

We cannot use "disable" or "deactivate"

paulpv:
> Just tested this : It does clean-up the frame, BUT this does NOT "disable" the source.
> 
> This behaviour can be good too, but we need to rename the feature.
> 
> Can be renamed :"Remove last received frame when NDI signal is lost" (or a better option)
> 
> We cannot use "disable" or "deactivate"

💯 % understood. I have tested this too.

I am going to move this back to Draft mode.

This is an experimental feature that, from my experience with Advanced Scene Switcher, was already expected to be a little weird.  
Even ASS says on their UI:  
image

And when you use ASS (without even using this PR code) it behaves "weird", according to their UI disclaimer, that the OBS UI does not show a source disabled or enabled.  
In this context the OBS UI can only show a "Scene Item" as disabled (hidden/invisible) or enabled (visible).

The semantics get very confusing here.  
Enabled/Disabled, Hidden/Shown, Visible/Invisible, Active/Inactive. 
"Source" vs "Scene Item"

It is a pretty big mess in OBS in general.

So I am still toying with this feature...  
... thus why I am putting back into Draft.

Trouffman:
Could this be tweak into a clear frame feature on connection loss ?
Also adding a signal, so that ASS can be the one controlling it (and free us from the burden).

Also this check is done for every loop, which on "low/slow" NDI sources or sources that experience small, intermittent lag, will cause the image to flicker or disappear/reappear. (Hard to reproduce but should be tested with some screen share on poor wifi)

BitRate27:
This enabled toggle looks to be tied to the scene item visibility UI in OBS. This can cause problems overriding the UI. For example, how do we know the difference between the user toggling it off (disabled) and us automatically disabling it when it goes offline? When the source comes back online, do we automatically enable it? Seems like they need a offline/online boolean on the source to implement this properly.

paulpv:
> This enabled toggle looks to be tied to the scene item visibility UI in OBS. This can cause problems overriding the UI. For example, how do we know the difference between the user toggling it off (disabled) and us automatically disabling it when it goes offline? When the source comes back online, do we automatically enable it? Seems like they need a offline/online boolean on the source to implement this properly.

Understood. It is unusable as implemented and needs a refactor. 
The current code should just be considered a rough proof of concept. 
It was originally AI generated, but I had to clean that up and it didn't have domain knowledge of what needs to actually be done. 
I am still poking at it, so it is still Draft now. 

I did not consult with anyone before I dove in and started to code this based on a user suggestion that was closely related to something similar that I was twiddling with in A.S.S.
Low priority for the team; just a personal interest of mine to hopefully help break me out of my personal "it's all pointless" mentality right now.
🧠

Trouffman:
@paulpv can we consider this managed by the latest #1330 ?

We should look to implement the signal handler as suggested by WarmUpTill https://github.com/DistroAV/DistroAV/issues/1316#issuecomment-3063394060

Trouffman:
As we have now implemented the #1276 .

Let's close this PR and work on a new PR to implement the signal_handler as suggested in : https://github.com/DistroAV/DistroAV/issues/1316#issuecomment-3063394060

PR #1318, opened 2025-07-08
buildspec: bump version to align with OBS 31.1

Trouffman:
Following closely the OBS upstream.

PR #1320, opened 2025-07-12
Update dependencies to 2025-07-11

Trouffman:
This PR only update the dependencies used by OBS.

This is to be used as a step toward using OBS 31.1 in the PR : #1318


Trouffman:
This PR is not longer needed.

Issue #1326, opened 2025-07-24
[Bug]: Main NDI Output Fails - WARN 426


jbcurler2010:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### OBS Log [URL] - Important!

https://obsproject.com/logs/3ri2XvsRBSBR9mo5

### Issue reported - Current Behavior

I can enable Main Output, close OBS, and re-open OBS and not have this issue. This only occurs if I enable Main Output, close OBS, and reboot the host. I have a Scheduled Task to start OBS (verbose log setting added). At this point, I should mention that our model for these systems has multiple Portable OBS instances starting via Scheduled Tasks at login. 

Note: NV12 is the selected (default) color format

Verbose Distro AV Log link above.

### Expected Behavior

Expect enabled Main Output to remain enabled on OBS launch under any condition.

### Steps To Reproduce

Set up Scheduled Task to Start Portable OBS instance at login (highest privilege and shutdown check disabled)
Open OBS
Enable Main Output
Close OBS
Reopen OBS
Verify Main Output is still enabled
Reboot Host
Check Main Output status

### Environment

```markdown
- OS: Windows 11
- OBS: 31.1.1 Portable
- DistroAV: 6.1.1 Portable
- NDI: 6.2.0.3
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [ ] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

BitRate27:
First thought is NDI is not able to create a sender because some background processes are not up and running yet. Might need to wait until NDI is ready to create senders before returning from library initialization functions. Or perhaps spawn a thread that waits for NDI to finish loading. 

I was able to duplicate the failure to create a sender when launching OBS as a scheduled task. I was not able to get the WARN-426 message. Also, my Main output was not disabled and OBS was outputting the Main output. 

I also tried to delay the start of OBS for 1 minute, and still got the failure to create a sender. 

This is what I am getting in the verbose log:

```
17:31:48.071: [distroav] WARN-416 - NDI Sender initialisation failed. 'OBS'
[...]

PR #1327, opened 2025-07-28
Check for sender not equal nullptr when creating output

BitRate27:
This addresses issue reported in https://github.com/DistroAV/DistroAV/issues/1326 where the format is reported as not supported. In actuality, the test for if the sender is created is not done properly and depending on the value of the ndi_sender, it can fail the if (o->ndi_sender) test, even when the valid sender is created.

As pointed out in the issue, the NDI documentation says this is a valid way to test (treating as a boolean), but is not adequate and need to be changed to suggest nullptr test instead.

### DIFF
--- a/src/ndi-output.cpp
+++ b/src/ndi-output.cpp
@@ -220,7 +220,7 @@ bool ndi_output_start(void *data)
 	send_desc.clock_audio = false;
 
 	o->ndi_sender = ndiLib->send_create(&send_desc);
-	if (o->ndi_sender) {
+	if (o->ndi_sender != nullptr) {
 		o->started = obs_output_begin_data_capture(o->output, flags);
 		if (o->started) {
 			obs_log(LOG_DEBUG, "'%s' ndi_output_start: ndi output started", name);


src/ndi-output.cpp
```diff
@@ -220,7 +220,7 @@ bool ndi_output_start(void *data)
 	send_desc.clock_audio = false;
 
 	o->ndi_sender = ndiLib->send_create(&send_desc);
-	if (o->ndi_sender) {
+	if (o->ndi_sender != nullptr) {
```
paulpv, line 223: Not sure why this makes any difference.

Unless I missed reading a log report somewhere, the code path seems to have always worked for non-null 99.999% use case.

For null/0/NULL/nullptr use case, we see it properly log `WARN-416 - NDI Sender initialisation failed` (SP).

So, not sure why changing `if (o->ndi_sender)` to `if (o->ndi_sender != nullptr)` would make any difference.

Are we ever seeing that `ndiLib->send_create` returns non-null, but `obs_output_begin_data_capture` returns false?
Even if so, I don't see how this change would make any difference.

BitRate27: I was getting the following in my Scheduled Task launch:

```
17:31:48.071: [distroav] WARN-416 - NDI Sender initialisation failed. 'OBS'
17:31:48.071: [distroav] 'OBS' ndi_output_start: ndi sender init failed
17:31:48.071: [distroav] -ndi_output_start(name='OBS', groups=''...)
17:31:48.071: [distroav] main_output_start: failed to start NDI Main Output 'OBS'; error=''
17:31:48.071: [distroav] ERR-400 - Failed to start NDI Main Output 'OBS'; error=''
```
But I was still getting output. This means a valid sender was getting created and returned, it was just that the boolean test was not checking correctly. When I changed it to compare against nullptr, then it stopped producing the warning 416 and error 400. Closest I come to an explanation is the pointer returned has the part that it looks at for a boolean is set to zero in my case. Only reason I thought this might fix the issue reported was that he was also getting the WARN-416. 

Now that we know multiple OBSs are being launched at the same time, I will try that and see if I can duplicate the format error also. Stay tuned...

BitRate27: After further investigation, I cannot explain why this change worked for my use case. I have tried smaller pointer values and cannot get the compiler to report a different outcome when using if (o->ndi_sender) and if (o->ndi_sender != nullptr). 
I will cancel this PR and address the original issue in another PR.

PR #1328, opened 2025-07-30
Add random number to NDI test output name

BitRate27:
When starting multiple OBS Studio instances at the same time, the creation of an NDI sender can fail if there is already a sender with the same name created. This PR adds a 5 digit random number to the name "NDI Output Support Test", so that every name is unique.

This is to address issue https://github.com/DistroAV/DistroAV/issues/1326 where the Main Output is disabled. This was because of "invalid format", but in reality the sender was not able to be created as the name already was being used by another instance of OBS. 

I tested this change in the same scenario as the one in issue 1326: 4 portable instances launched at the same time in Task Scheduler. I confirmed all main outputs were started and log files were clean or Errors and Warnings.

Also in this PR is NOT turning off the Main Output in the config saved for the user. This does not change the behavior for the original case where a user has chosen an unsupported output format. 

This was tested by:
1. Toggling off the main output, changing the pixel format, and then not being able to turn on main output.
2. With main output turned on, loaded a profile with an invalid pixel format and it did not crash. When bringing up the output dialog, the main output was still turned on, but the container was not enabled and not outputting main. 
3. When choosing a profile with a supported pixel format, the main output restarted as it was still turned on.

### DIFF
--- a/src/main-output.cpp
+++ b/src/main-output.cpp
@@ -18,6 +18,7 @@
 #include "main-output.h"
 
 #include "plugin-main.h"
+#include <random>
 
 struct main_output {
 	bool is_running;
@@ -109,7 +110,17 @@ bool main_output_is_supported()
 	auto output_groups = config->OutputGroups;
 
 	obs_data_t *output_settings = obs_data_create();
-	obs_data_set_string(output_settings, "ndi_name", "NDI Output Support Test");
+
+	// Append a random number to the test NDI name to avoid
+	// conflicts with existing outputs
+	std::random_device rd;
+	std::mt19937 gen(rd());
+	std::uniform_int_distribution<> dis(10000, 99999);
+
+	int randomNum = dis(gen);
+	std::string testString = "NDI Output Support Test " + std::to_string(randomNum);
+
+	obs_data_set_string(output_settings, "ndi_name", testString.c_str());
 	obs_data_set_string(output_settings, "ndi_groups", "DistroAV Config");
 
 	bool is_supported = true;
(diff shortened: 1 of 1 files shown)


src/main-output.cpp
```diff
@@ -109,7 +110,17 @@ bool main_output_is_supported()
...
-	obs_data_set_string(output_settings, "ndi_name", "NDI Output Support Test");
+
+	// Append a random number to the test NDI name to avoid
+	// conflicts with existing outputs
+	std::random_device rd;
+	std::mt19937 gen(rd());
+	std::uniform_int_distribution<> dis(10000, 99999);
+
+	int randomNum = dis(gen);
+	std::string testString = "NDI Output Support Test " + std::to_string(randomNum);
```
paulpv, line 121: This is a nitpick...

I agree that `std::string` is "better" than `QString`, but our code almost exclusively uses QString almost everywhere, so I would prefer to be consistent.

https://github.com/search?q=repo%3ADistroAV%2FDistroAV+QString&type=code
98 uses.

https://github.com/search?q=repo%3ADistroAV%2FDistroAV%20std%3A%3Astring&type=code
37 uses.
28 of them in `obs-support`, which is technically not our code (copied from OBS internal/private code).
5 of them in our `ndi-finder` code that I guess I let slip through the cracks.
1 that calls that `obs-support` code.
1 that calls our `ndi-finder` code.
1 a map in `config.cpp` that I guess I also let slip through the cracks.
1 a map in `ndi-source.cpp` that I guess I also let slip through the cracks.

Trouffman (commented):
To avoid confusion and be easier to maintain. Please use a more descriptive variable name.
Some idea : ndiOutputSupportName

BitRate27:
I can change the variable name so it is more descriptive. Has anyone built/downloaded and tested this yet?

BitRate27:
Use another branch. Mistakenly included debug changes on original branch for this PR.

PR #1330, opened 2025-08-04
Implement "Keep Last Received Frame" functionality

Lordmau5:
This is a new checkbox that is enabled by default. When it is enabled it will keep the last frame that was received on the source, no matter if the sender NDI instance was disconnected or is sending "none" frame data. When it is disabled, it will wait for up to 3 seconds to receive a new frame and, if there is no new frame, will clear it on the receiver end.

The way this is possible is via. a new helper method called `is_filter_invalid` (probably not the final name) which is checking the source width and height, as well as if the filter is enabled on the source and if the source is active anywhere. Sources don't update when they aren't rendered in the main OBS view from what I can tell (e.g. a Browser Source would just seem like it's frozen) so this is in-line with the current way OBS handles it.

This PR supersedes #1276 by for example making sure the NDI connection stays alive instead of destroying it the moment the source / filter gets invalid.


src/ndi-filter.cpp
```diff
@@ -65,6 +65,8 @@ const char *ndi_audiofilter_getname(void *)
 }
 
 void ndi_filter_update(void *data, obs_data_t *settings);
+void destroy_ndi_sender(ndi_filter_t *filter);
+void destroy_and_create_ndi_sender(ndi_filter_t *filter, obs_data_t *settings);
```
Trouffman, line 69: nitpicking here : Naming is expected to stay consistent with other (output) logic and code in this file, can this be only "ndi_sender_create". and `destroy_ndi_sender` be `ndi_sender_destroy`.

Even if the logic inside that funciton "will check if it needs to be destroyed first".

src/ndi-filter.cpp
```diff
@@ -107,24 +109,52 @@ void ndi_filter_getdefaults(obs_data_t *defaults)
 	obs_log(LOG_DEBUG, "-ndi_filter_getdefaults(...)");
 }
 
+bool is_filter_invalid(ndi_filter_t *filter)
```
Trouffman, line 112: Nitpicking (level 999).

Flagging a state should be done with "positive" in mind : filter_is_valid. rather than negative (is_invalid). 

Reasoning :
Having a "negative" function returning "true" is usually confusing (especially on multi-criteria functions) when a "nagtive" state is expected.

src/ndi-filter.cpp
Lordmau5: An alternative I could propose would bet o add an extra argument to the `ndi_create_sender` method that's like `force_recreate` or something, would that be better?

Lordmau5:
Okay I've renamed `is_filter_invalid` to `is_filter_valid` with inverted logic.

Additionally, I thought a bit more about the `ndi_sender_create` method and its naming and I think it *is* better that we just name it that and make sure it destroys the sender during it. 
It's only really called with `ndi_filter_update` (when the filter updates) and in `ndi_filter_tick` (during video ticks, and only if `f->ndi_sender` is a nullptr)

BitRate27:
In addition to allowing a NDI source to go black if no NDI data is received after 3 seconds, this PR also allows a second instance of OBS receiving Dedicated NDI Output from another OBS to also go black when no data is being sent.

Prerequisites:

Two portable OBSs (OBS 1 and OBS 2) running
Test Pattern App from NDI Tools running

1. Open OBS 1
2. Add a new NDI Source using
3. Select Test Pattern as NDI source
4. Uncheck "Keep the last frame received"
5. Add Dedicated NDI Output filter to NDI Source
6. Open OBS 2
7. Add a NDI Source
8. Select Dedicated NDI Output as NDI source
9. Uncheck "Keep the last frame received"
10. Shutdown Test Pattern App
- Test Pattern disappears in OBS 1 in 3 seconds
- Dedicated NDI Output disappears in OBS 2 in 6 seconds

Observations:
OBS is responding correctly to the Test Pattern NDI Source terminating.
OBS is responding correctly to a Dedicated NDI Output sending an empty frame (due to invalid frame on NDI source hosting filter)
Dedicated NDI Output still exists after filter disabled or source no longer sending (just nothing being sent)

Trouffman (commented):
While reviewing this change, the naming triggered me multiple time (especially as we use this "last frame" codename for the "behavior".

After some thought I suggest the following :
- Behavior is a "Connection Behavior"

Also having only a tick box is lacking context and might lead to confusion in the same manner we had when developing the "behavior" feature.

Suggestion :

"On timeout"
- Keep last received content (frame)
- Clear/reset the source (3sec)

PoC : 
image

Lordmau5:
> While reviewing this change, the naming triggered me multiple time (especially as we use this "last frame" codename for the "behavior".
> 
> After some thought I suggest the following :
> 
>     * Behavior is a "Connection Behavior"
> 
> 
> Also having only a tick box is lacking context and might lead to confusion in the same manner we had when developing the "behavior" feature.
> 
> Suggestion :
> 
> "On timeout"
> 
>     * Keep last received content (frame)
> 
>     * Clear/reset the source (3sec)
> 
> 
> PoC : image

Much better than the checkbox, I agree - I'm not good with UI/UX and while it made sense in *my head* I can fully see it causing confusion for other users

Trouffman:
This should be feature-complete. Translation added in all language using machine learning except for : English, French, Spanish

Trouffman (approved):
Reworked the tickbox to be a drop-down for better clarity.

PR #1332, opened 2025-08-07
Add random number to NDI test name

BitRate27:
When starting multiple OBS Studio instances at the same time, the creation of an NDI sender for testing for supported pixel format can fail if there is already a sender with the same name created. This PR adds a 5 digit random number to the name "NDI Output Support Test", so that every name is unique.

This is to address issue [1326](https://github.com/DistroAV/DistroAV/issues/1326) where the Main Output is disabled. This was because of "invalid format", but in reality the sender was not able to be created as the name already was being used by another instance of OBS.

I tested this change in the same scenario as the one in issue [1326](https://github.com/DistroAV/DistroAV/issues/1326): 4 portable instances launched at the same time in Task Scheduler. I confirmed all main outputs were started and log files were clean of Errors and Warnings.

Also in this PR is NOT turning off the Main Output in the config saved for the user. This does not change the behavior for the original case where a user has chosen an unsupported output format.

This was tested by:

1. Toggling off the main output, changing the pixel format, and then not being able to turn on main output.
2. With main output turned on, loaded a profile with an invalid pixel format and it did not crash. When bringing up the output dialog, the main output was still turned on, but the container was not enabled and not outputting main.
3. When choosing a profile with a supported pixel format, the main output restarted as it was still turned on.

Note: Previous branch used for previous PR to address this issue got corrupted. This PR includes changes suggested by @paulpv and @Trouffman in that PR.

### DIFF
--- a/src/main-output.cpp
+++ b/src/main-output.cpp
@@ -18,6 +18,7 @@
 #include "main-output.h"
 
 #include "plugin-main.h"
+#include <random>
 
 struct main_output {
 	bool is_running;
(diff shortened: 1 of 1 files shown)


Trouffman (approved):

BitRate27:
```
ey, I appreciate your help
I don't believe I had/have multiple instances of OBS running. If I were to open another instance, I am met with the warning popup.
Is it possible that something else is causing the conflict?

Either way, the build you provided did resolve the setting issue, and I can close and reopen OBS to my heart's content without re-checking that box for the time being.
Thank you! (edited)
```
Had a user verify the change works.

Issue #1271, opened 2025-05-12
[Bug]: crash report


walker-WSH:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### Issue reported - Current Behavior

- WIN 11
- OBS 31
- NDI 6.0.0
- 

Steps (100% crash)

- run obs, open ndi UI by Tools menu
- click "check for update" in ndi dialog
- click "cancel" soon in update dialog of ndi
- quikly click "check for update" more thank 10 times
- close ndi dialog, OBS crashed

callstack of NDI 6.0.0
<img width="653" alt="Image" src="https://github.com/user-attachments/assets/95068438-7f62-4cff-84d9-f314649ab524" />

### Expected Behavior

obs no crash

### Steps To Reproduce

as above

### Environment

```markdown
- OS:WIN11
- OBS: 31
- DistroAV: 6.0
- NDI:
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [x] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

RayneYoruka:
May I ask.. why did you click for 10 times on the check for updates? It makes no sense.

paulpv:
Don't "quikly click "check for update" more thank 10 times"

paulpv:
The crash is probably just because the settings dialog is closed when an update completes. 

Low priority unless lots of people run into this. 

I once tried to add Firebase Crash Reporting to DistroAV to track things like this; was never able to get it to work.

walker-WSH:
hi   @RayneYoruka  @paulpv 

Sorry for the late response.
Oh~~ This crash is reported by users. Of course, clicking in rapid succession is pointless.

Trouffman:
This issue is reported while using PRISM Studio.

Please reproduce with OBS and provide the log file. (even better if launching obs with --distroav-debug --verbose)

walker-WSH:
> This issue is reported while using PRISM Studio.
> 
> Please reproduce with OBS and provide the log file. (even better if launching obs with --distroav-debug --verbose)
> 

hi @Trouffman 

[...]

PR #1339, opened 2025-08-16
#1271 UI/Settings: Ensure update dialog is alive

walker-WSH:
Fix a crash in https://github.com/DistroAV/DistroAV/issues/1271

Moderator edit :
- This fix a crash when the dialog is closed before the URL response is received.


Trouffman (approved):
Tested by @BitRate27.
Simple fix.

PR #1341, opened 2025-08-17
Uniformise all translation files. (some machine translated, some not)

Trouffman:
This PR address 2 problems :
- The options for the Behavior parameters where not populated in the translated languages.

- Uniformise & Order all the language files with the same structure.

Additionally : 
- Machine translated some of the fields

Ideally :
- Users should suggest correction in their native language and if a language cannot be maintained it should be retired.


Trouffman:
For context, it is FINE to not have all the field translated, and if this is the case, they should not be defined in each language and will fallback to the "default" language (en-US)

Trouffman:
All translations will now be managed at Crowdin : https://crowdin.com/project/distroav

PR #1347, opened 2025-08-28
Fix preview tally

BitRate27:
This PR fixes the issue brought up in [NDI Tally Light Issues...](https://github.com/DistroAV/DistroAV/issues/687).

In summary, when a source is in program, and not in preview, the preview tally is still sent to the camera. This appears as red and green bars at the top of NDI Studio Monitor, or a camera tally light flashing red/green as reported in the original issue above.

This PR also allows the preview tally to reflect the state of the visible setting of a source's scene item.

The cause of the issues is that the show/hidden callbacks and state retrieval methods in OBS, do not reflect the preview state of sources. 

To try to properly fix the issue, in the OBS PR [Allow sources to know when it is being previewed](https://github.com/obsproject/obs-studio/pull/11898), preview/depreview callbacks were added to OBS, similar to the activate/deactivate/hide/show callbacks. Since the OBS PR has been sitting for 6 months and comments in the PR suggest the risk is too great to change OBS at this level, this PR implements the preview/depreview APIs in obs-support/obs-source-preview.cpp.  This solution provides a simple way for the NDI source to implement preview tally correctly, by merely using the preview/depreview APIs instead of show/hide.

The majority of the code is in the obs-support/obs-source-preview.cpp which uses Frontend events and scene item signals to maintain the preview state of a NDI source.

The visibility of a source in the preview scene can be altered in the following ways:
- Toggle visibility in preview scene
- Select/swap scene for preview
- Source in a group (group visibility has priority)
- Move source to group
- Change NDI source in source properties
- Load a new scene collection

All these events/signals should cause a new tally state to be sent to the camera.

As an added complication, OBS scene item APIs, specifically the signals, do not operate as expected in the following ways:
- When a source is in a group, the item_visible doesn't get fired when it is toggled by the user
- Sources in a group need to connect to the group's item_visible signal
- When a group is toggled, the item_visible doesn't get fired for the items in the group

A good part of the code in obs-source-preview is to deal with these deficiencies.

This PR was tested by:

- Sending 3 separate test patterns using NDI Test Patterns App.
- Opening up 3 NDI Studio Monitor Apps, for each of the test patterns
- Turning on Tally Overlay in the monitors
- Creating 3 NDI sources, one for each pattern 

Once setup, testing was done by altering the visibility of the NDI sources as listed above and verifying the tallys shown in NDI Studio Monitor are correct.

image


Trouffman:
@BitRate27 last part for this would be to have a new "deterministic" documentation :
- How it is expected to behave in GENERAL (aka the overall guideline to define if. Behaviour is a bug or not)
- When a Tally status is expected to change

Thanks !

Issue #1171, opened 2024-12-22
[Tweak]: Only read the installGUID an never write it to avoid possible conflict in App config file


Trouffman:
### Pre-Report Checks

- [X] This is not a Bug Report
- [X] This is not a Feature Request
- [X] This is not a Support Request
- [X] This could not be reported anywhere else (Issue/discussion)

### Description of your report

Looking at the codebase the plugin use for the update. One of the required element is the "installGUID", this is the same used by the OBS Project.

This is actively set by OBS this means : 
- Realistically, it would never be set by the plugin. -> Making that part of the codebase "redundant"
- If, for any reason, the plugin does write it, it is writing this in the General part of the config file AND could conflict (none report so far) with the way OBS Core is already doing.

The proposed changes :
- The plugin should only "Read the InstallGUID from the config file"
- If the installGUID is not set, assume the user does not want to have updates.

Expected avantages
- Simplify codebase
- Avoid maintaining non-used code
- Reduce possible conflicts with OBS Core
- Simplify Update mechanism

Test done already 
- Confirmed by testing that the installGUID is always created (by OBS), even when the OBS updates are turned off in UI & via config file.

### I could not report this anywhere else because...

It is a future-looking improvement, not a bug per say.

Trouffman:
@paulpv this touches some of your code.

If we write something, let's be in a field that is NOT controlled/used by the upstream obs-studio.

We could have this UUID as part of the [NDI/DISTROAV] Section of the global Config file.

We could easily port it to keep that same ID for those that have one, or create one for those that are missing (and first-launch OBS).

PR #1349, opened 2025-08-29
Update-Process - Change the way the GUID is expected to exist

Trouffman:
This PR should address the challenge set on #1171 

This would effectively read the GUID set by OBS and never write it.

We will need a way to detect a new "install" maybe with a placeholder value for the update check.

This is because on a FIRST clean OBS launch, the GUID does not exist until there is a "save" event in OBS (settings change or OBS close).

This means that when loading the plugin, that GUID does not exist yet.

The Plugin should never write this because :
- This is not the plugin configuration space (aka do not mess with others' settings, especially not OBS)
- This could be used/read but never created or modified.

This can be addressed by :
A. Only read the GUID if it exists, use a fallback & identifiable value if not.

B. Create our own GUID in our own config space.

This PR implement option A.


Trouffman:
The process now only "read" the Install GUID from OBS "General".

If no GUID is detected (meaning that OBS is launched for the first time) this will returned a dummy identifiable GUID (needed for the update process)

This is set to `noguiddetectednewfreshinstalldistroavobs` for now, it's 40 characters long like the expected GUID.

image

This is also managed in the log entries to be easier to debug :
`[distroav] No OBS GUID detected - Assuming this is a new installation`

This uses the OBS GUID as soon as it exists (aka when a "save" event occurs.)

Trouffman:
@paulpv please review this change as it update the behavior you initially put in place.

Trouffman:
Note : 
We now only use `CalculateFileHash()` from `shared-update` maybe this could be merge into the update.cpp to simplify code / files / deps

PR #1350, opened 2025-09-05
RFP: Refactor transport layer into core and NDI modules

paulpv:
## Summary
- introduce core transport interface and global accessors
- add NDI transport implementation and factory
- move NDI-specific code into dedicated transport library
- update output settings and plugin entry to use transport interface
- restructure CMake to build core and NDI libraries

## Testing
- `cmake --preset ubuntu-x86_64` *(fails: Could not find a package configuration file provided by "LibObs")*

------
https://chatgpt.com/codex/tasks/task_e_68ba4462a17483339bc61aaa62c8b5e0


BitRate27:
After my first visual run through, the concept looks sound and will easily support other sources and outputs. Only minor comment right now is the choice of 'av' for the namespace. What is the reason for adding the namespace? If needed, maybe 'transport' would be better. Still looking at the PR, but looks good so far.

BitRate27:
Since we can have a different obs_frontend_add_event_callback for each transport. Can we move that callback to the NDITransport::initialize so it stays with the NDI code?

(Can you show me how to comment in the code like you did?)

paulpv:
> (Can you show me how to comment in the code like you did?)

Hover on or around a line number and a "+" sign will appear to comment that line

BitRate27:
I went through and created a new source/transport with no output capabilities. Created the *Transport.{hpp,cpp} files and the sub dir in transports folder. Also added the lines in the CMakeLists.txt file. I know it is a draft, as it doesn't build rn, but I was able to see what needs to be done, which was helpful. I didn't see that much in common to justify the change, but I had no output. It does clean up the plugin-main.cpp code quite a bit and organizing into sub libraries does help with the encapsulation. 

One thing I think needs a re-look is the factory creation part. Would be nice to have a registration process. And a glaring issue, which may be because I don't completely understand the factory concept, is how to get the Transport you want. The set_transport/get_transport makes it awkward to work with more than one transport at the same time, like in the Output Settings dialog. Would rather just have a way to get the specific transport I want.

paulpv:
The current code would be a stepping stone to just refactoring the code to abstract out all NDI stuff; It would still only ever produce a single plug-in that supports NDI.
The next step would be even more of a refractor to then make to plug-in agnostic and a transport/protocol/brandless library that can be told a single or multiple transport/protocol/brand to use. 
Next step beyond that would be to build a concrete branded instance that is the actual plug-in distributable itself that specifies what transports/protocols/brands to use:
Examples: 
DistroAV NDI
DistroAV NDI|HX
DistroAV NDI|HX3
DistroAV OMT (our special non-C# spin on OMT's OBS plug-in)
DistroAV WebRTC (our special/better spin on something OBS can already do)
DistroAV Pro (All)
Any combination/permutation of the above, multiple

PR #1351, opened 2025-09-05
RFP: feat: allow selecting audio tracks for NDI output

paulpv:
## Summary
- let users pick which audio tracks are included in NDI output
- expose track selection in global output settings
- persist selected tracks in config

## Testing
- `cmake -S . -B build` *(fails: Could not find a package configuration file provided by "LibObs")*
- `./build-aux/run-clang-format src/config.h src/config.cpp src/forms/output-settings.cpp src/forms/output-settings.ui src/main-output.cpp src/ndi-output.cpp data/locale/en-US.ini` *(fails: /usr/bin/env: 'zsh': No such file or directory)*

------
https://chatgpt.com/codex/tasks/task_e_68ba503006548333a5e7c86b8185ad7d


Trouffman:
There are quite a lot of changes that seems not changes at all.

Might need to be looked at what codex is trying to do there and have a clean PR.

The bitwise approach seems good for this goal.

PR #1359, opened 2025-09-14
Add new OBS source to receive WHIP output

BitRate27:
This PR adds a new source called WHIP to OBS. The main goal of the source is to support low bandwidth input (H.264) to OBS from another WHIP sender (eg. OBS streaming output option).

The initial implementation works between two OBS instances on the same machine. 

To Test:

1. Start a version of OBS that supports WHIP streaming output: >= 30.0.0
2. Add a video input source (NDI, Media Source, Window Capture,...).
3. On Settings->Stream, select WHIP output. Select OK for Opus audio output.
4. Enter URL as: http://localhost:8080/whip
5. On Settings->Video, set scaled output to 1280x720.
6. Start a version of OBS with the DistroAV plugin installed (see artifacts for this PR)
7. Add a WHIP source to the scene.
8. Enter URL as /whip
9. Select OK
10. Go to OBS started in step one and select "Start streaming".

Priority of ToDos:

1. Work across different machines.
2. Build on mac and linux.
3. Recognize resolution of sender.
4. Show Bearer token and use it.
5...

This is a work in progress...

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ brew install --cask distroav
 Linux ([Flatpak](https://flatpak.org/)) ![Flathub Version](https://img.shields.io/flathub/v/com.obsproject.Studio.Plugin.DistroAV?link=https%3A%2F%2Fflathub.org%2Fapps%2Fcom.obsproject.Studio.Plugin.DistroAV)
 ```
 flatpak install com.obsproject.Studio com.obsproject.Studio.Plugin.DistroAV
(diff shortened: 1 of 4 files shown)


edwardedwards99-a11y (approved):

SNRSE:
This feels like it should be in a separate plugin. I like the idea of this PR of adding a WHIP server. But that is not something I would expect within DistroAV

PR #1367, opened 2025-10-04
Plugin Manager Name string

BarryCarlyon:
Tweak to use a more useful name for the Plugin Name for Plugin Manager Purposes.

Apply to all lang files for future translation where relevant

image


BitRate27:
Thanks for volunteering to update this. If we are going to change the name from 'distroav' to something more appealing, then I suggest just 'DistroAV'. NDI ties us to only NDI and a big part of the name change we went through last year was to remove OBS and NDI from the name. Also, we have clearance from NDI that we only need to use the (r) on the main Output Settings page and not include it anywhere else.
Regarding adding it to the translations, since it is a product name and not a parameter or setting, it will be DistroAV in any language.

BarryCarlyon:
Reverted "use of translation" (looks like I stipped a space on one of the lang files which is by the by)

src/plugin-main.cpp
```diff
@@ -39,7 +39,7 @@ OBS_MODULE_USE_DEFAULT_LOCALE(PLUGIN_NAME, "en-US")
 
 const char *obs_module_name()
 {
-	return PLUGIN_NAME;
+	return "DistroAV"
```
paulpv, line 42: Needs to be `PLUGIN_DISPLAY_NAME`

src/plugin-main.cpp
```diff
@@ -109,7 +109,7 @@ QString makeLink(const char *url, const char *text)
  *  5. The message box is shown after a delay (default 2000ms)
  *  6. Shows the dialog as WindowStaysOnTopHint and NonModal
  *  7. Deletes the dialog when closed
- *
```
paulpv, line 112: Revert

src/plugin-main.cpp
paulpv, line 42: Needs to be `PLUGIN_DISPLAY_NAME`

Trouffman:
image
Success!

PR #1375, opened 2025-11-11
Source: update Dutch translation file

0xCipherShard:
Update the Dutch translation (`nl-NL.ini`) to include new strings added in recent updates and correct several minor translation errors and outdated phrases.


Trouffman:
Thnak you for the update.

Issue #1335, opened 2025-08-11
[Bug]: audio_thread created by ndi-6.1.1 is not closed during exiting


walker-WSH:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [ ] I have checked that my OS / OBS / NDI versions are supported

### Issue reported - Current Behavior

This crash is found at user's side with PRISM exiting. I do not find the reproduce steps.
the `audio_thread` created by ndi (6.1.1) is not closed during exiting.

> **1. obs_shutdown has been called**

<img width="834" height="159" alt="Image" src="https://github.com/user-attachments/assets/f8d773c6-2908-4e0b-bfd2-e13fb5a3912d" />

> **2. However, the audio_output in libobs (0X1FE37C1D540) created by 3rd plugin is not exited**

<img width="1187" height="229" alt="Image" src="https://github.com/user-attachments/assets/c90245b4-7917-47a4-9071-b2931e87c3bd" />

> **3. we can find this audio_output is set for this output (0X1FE37CF56A0)**

<img width="1176" height="90" alt="Image" src="https://github.com/user-attachments/assets/5207c0de-8042-4f8e-b594-b2a1125b5c72" />

> **4. this output (0X1FE37CF56A0) is created by ndi plugin**

<img width="1472" height="664" alt="Image" src="https://github.com/user-attachments/assets/fb7a88a1-3ac8-4557-a68e-8d0b2795335a" />

### Expected Behavior

during exiting, the created audio_output should be closed

### Steps To Reproduce

I have not reproduced it.
This crash is happened at the user's with PRISM Live Studio

### Environment

```markdown
- OS:  Windows 11(x64)
- OBS:  PRISM 5.0.0.635
- DistroAV: 6.1.1

This issue was modified in the past
https://github.com/DistroAV/DistroAV/pull/996

I find `audio_output_close` is only called in `preview_output_stop`. 
`preview_output_stop` is called from `OBS_FRONTEND_EVENT_EXIT》preview_output_deinit`

That front event is also sent from PRISM as obs's code.
```

### Pre-Check for Bug Report

- [ ] I am able to reproduce the bug following the steps provided above
- [ ] I have confirmed this bug exist on more than one device/computer
- [ ] I have tested using a new scene collection to reproduce the issue
[...]

PR #1377, opened 2025-11-13
Check preview output instead of is_running to prevent leaks

BitRate27:
This fixes https://github.com/DistroAV/DistroAV/issues/1335 by checking for the existence of preview output instead of is_running.   The is_running check was causing a leak of the audio thread. 

This can be duplicated by launching an OBS and turning on preview output in DistroAV Output Settings. In another OBS, also turn on Preview Output with the same NDI name as the first OBS. This does not create a sender because one already exists with the same name. When the second OBS Preview Output, which was not created, is turned off, it doesn't close the preview output because it is not actually running. The output was created, it just wasn't running. 

The fix is to check if the output exists instead of checking if it is running.


Trouffman:
Should we update this on the main_output ?

Should we revise the is_running for this purpose everywhere too?

BitRate27:
When I did this I verified the issue did not exist on Main. However, for the sake of consistency, I went ahead and dug into the differences between starting and stopping main and preview output and found that preview did not report the error and stop the output if it was not able to create the ndi output when the name was already used (by another instance of OBS). So I added code to stop the preview output if it was not created. I also checked context.output instead of context.is_running when stopping main output.

FYI: Filter output does not use the obs_output pipeline, no audio thread started, so nothing to do in that part of the code.

BitRate27 (commented):
Have confirmed the audio thread does not leak when creating preview output with the same name as and existing output (ie. main). Also added the code recommended in the issue discussion to bdupstr the ndi_name for OBS voi and aoi info as suggested in the original [issue](https://github.com/DistroAV/DistroAV/issues/1335). This is just to be correct and not cause any doubt when finding issues with output from DistroAV.

PR #1384, opened 2025-12-17
Send user to NDI Tools download instead of runtime

BitRate27:
When DistroAV cannot find the NDI libraries, we put up an error dialog and provide a link to download the runtime. The link to the NDI runtime (http://ndi.link/NDIRedistV6) provided in the NDI SDK include file does not work as of 12/17/2025 and hasn't for a while. While we wait for NDI to address the issue, we now send the user to the web page to Download NDI Tools for their platform: https://ndi.video/tools/download

To do this, we use the already existing NDI_OFFICIAL_TOOLS_URL symbol set in plugin-main.h

### DIFF
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -280,7 +280,12 @@ bool obs_module_load(void)
 	if (!ndiLib) {
 		auto title = Str("NDIPlugin.LibError.Title");
 		auto message = "Error-401: " + QTStr("NDIPlugin.LibError.Message") + "<br>";
+
+#if defined(NDI_OFFICIAL_TOOLS_URL)
 		message += makeLink(NDI_OFFICIAL_TOOLS_URL);
+#else
+		message += makeLink(NDI_OFFICIAL_WEB_URL);
+#endif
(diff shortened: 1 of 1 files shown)


Trouffman:
@BitRate27 the link does work. There might have been some issues in the past and on some environment that the http:// link does not work if the redirect is a strict to HTTPS.

This issue with the short link has been reported to NDI multiple times, last update was that it "has been fixed".

NB: If this happens again, we will use our own redirect link and play with that instead to not break the User-experience.

BitRate27:
I think the real issue here is that the link provided in the DistroAV dialog when the Runtime is not found is problematic for new users. In most cases clicking on the link will download a the Runtime exe. This download is sometimes hidden, based on browser or preferences. This can lead users to deciding the link does not work (happens to me every time, and I have done it many times and still get tripped up.) Though installing the Runtime has a smaller footprint on the user's machine than installing the Tools, this is the first step and they are not being exposed to the fact that DistroAV has a major dependence on a third party. Pointing them to the NDI website, they are more aware of the dependency DistroAV has on the third party. I feel linking to the NDI Tools is a better user experience than a "hidden" download.

RayneYoruka:
I agree that there can be some confusion with new users, not to mention the issues that tend to happen with the NDI links not working.

Issue #1386, opened 2025-12-23
[Bug]: Audio/Video Sync "Network" Mode Fails to Use Timestamps When OBS Starts Before NDI Sources, Leading to Dropped Frames


MarekPro1:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### OBS Log [URL] - Important!

https://obsproject.com/logs/BXKechPRUSVPxcjl

### Issue reported - Current Behavior

When using the DistroAV plugin in OBS Studio with NDI sources (e.g., an NDI camera), and both the OBS host and NDI sources are synchronized via NTP, the "Network" audio/video sync setting is intended to display frames based on their timestamps. However, if OBS is started before the NDI sources become available, the plugin appears to fallback to displaying frames in the order they arrive rather than by timestamp. This results in out-of-sync playback and frame drops.

lagging log file: https://obsproject.com/logs/BXKechPRUSVPxcjl
smooth log file: https://obsproject.com/logs/jFHm4UQDQbOUD1XZ

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/c114f1ed-9bee-4953-b24a-19669d24a70c" />

### Expected Behavior

Expected Behavior:
With "Network" sync enabled and NTP synchronization in place, frames should be displayed strictly according to their timestamps, regardless of when OBS is started relative to the NDI sources. This should prevent frame drops and maintain proper synchronization.
Actual Behavior:
If OBS starts before the NDI sources, frames are displayed in the arrival order instead of by timestamp. This causes desynchronization and leads to dropped frames, as the timing information is ignored.

### Steps To Reproduce

1. Ensure the Windows computer running OBS and the NDI camera/source are time-synchronized using NTP (e.g., via a shared NTP server).
2. In OBS, add an NDI source and set the "Sync" option to "Network" under the source properties.
3. Start OBS Studio before powering on or making the NDI camera/source available on the network.
4. Once the NDI source becomes available, observe the video playback in OBS.

### Environment

```markdown
- OS: Windows 11 IoT Enterprise LTSC 24H2
- OBS: 32.0.2 installed from .exe
- DistroAV: 6.1.1
- NDI: 6.2.1.0
```

### Pre-Check for Bug Report

[...]

PR #1388, opened 2025-12-27
Fix Network based Audio/Video Sync

BitRate27:
This fixes a problem we have with the Audio/Video Sync setting of Network.

Problem:
When Network is chosen, the following OBS verbose log messages will get logged...
```
16:23:03.271: Timestamp for source 'NDI Source 2' jumped by '3299999967', expected value 10664201107917214989, input value 10664201111217214956
16:23:03.305: Timestamp for source 'NDI Source 2' jumped by '3299999967', expected value 10664201111250548289, input value 10664201114550548256
16:23:03.337: Timestamp for source 'NDI Source 2' jumped by '3299999967', expected value 10664201114583881589, input value 10664201117883881556
```
This is because the timestamp values of the NDI source is much bigger than expected. In our case the unit of the timestamp is 100th of a nanosecond.

To fix this, we do not need to multiply the timestamp from NDI by 100 when passing on to OBS. It is not clear when this was changed as I do remember the code only multiplied the timecode by 100. 

This is an attempt to fix issue https://github.com/DistroAV/DistroAV/issues/1386 where a user is having issues using NTP on his cameras and in DistroAV.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -758,7 +758,7 @@ void ndi_source_thread_process_audio3(ndi_source_config_t *config, NDIlib_audio_
 
 	switch (config->sync_mode) {
 	case PROP_SYNC_NDI_TIMESTAMP:
-		obs_audio_frame->timestamp = (uint64_t)(ndi_audio_frame->timestamp * 100);
+		obs_audio_frame->timestamp = ndi_audio_frame->timestamp;
 		break;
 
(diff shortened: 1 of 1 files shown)


BitRate27:
According to the specification for the NDI receiver API, timestamp is 100 nanosecond increment, just like timecode. 

Closing this PR as the current implementation is correct. It needs to be multiplied by 100 before sending to OBS (nanoseconds).

PR #1398, opened 2026-01-29
Update NDI includes + request 6.3 minimum NDI version

Trouffman:
This update to the latest SDK and request/enforce the use of NDI Runtime 6.3, in prevision of using the Discovery capabilities inside the plugin. 

Expected gain from this upgrade:
- Enable the discovery features in upcoming codebase
- Performance improvement (from new NDI SDK)

### DIFF
--- a/CI/libndi-package.sh
+++ b/CI/libndi-package.sh
@@ -6,7 +6,7 @@
 
 set -e
 
-LIBNDI_VERSION="6.2.0"
+LIBNDI_VERSION="6.3.0.3"
 
 SCRIPT_DIR=$(dirname "$0")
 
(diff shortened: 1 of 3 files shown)


BitRate27:
On Windows, with NDI 6.2 installed, I installed the version of DistroAV in the Assets of this PR.
When I ran OBS, it said that I need to have NDI 6.3 installed. I clicked on the link in the dialog and downloaded the new Runtime exe. I shut down OBS, executed the NDI Runtime.exe and then re-opened OBS. It no longer complained about needing 6.3 installed.

Next I will fetch the PR source and build in Visual Studio to confirm that developer part works.

PR #1399, opened 2026-02-08
Start main output and preview after other plugins have setup output

exeldro:
Start main output and preview queued on the UI thread to allow other plugins to configure their canvases

Aitum Stream Suite fails to configure its canvases because there is an output active. The configuration of the canvases is done on OBS_FRONTEND_EVENT_FINISHED_LOADING in Aitum Stream Suite, but OBS_FRONTEND_EVENT_FINISHED_LOADING is called on DistroAV before that.

### DIFF
--- a/src/plugin-main.cpp
+++ b/src/plugin-main.cpp
@@ -360,8 +360,14 @@ bool obs_module_load(void)
 		obs_frontend_add_event_callback(
 			[](enum obs_frontend_event event, void *) {
 				if (event == OBS_FRONTEND_EVENT_FINISHED_LOADING) {
-					main_output_init();
-					preview_output_init();
+					auto main_window = static_cast<QMainWindow *>(obs_frontend_get_main_window());
+					QMetaObject::invokeMethod(
+						main_window,
+						[] {
+							main_output_init();
+							preview_output_init();
+						},
+						Qt::QueuedConnection);
 				} else if (event == OBS_FRONTEND_EVENT_EXIT) {
 					// Unknown why putting this in obs_module_unload causes a crash when closing OBS
 					main_output_deinit();


BitRate27:
Could you list the steps needed to reproduce this issue in Aitum Stream Suite. I have downloaded and added a recording output, but have no problem loading the Aitum plugin with NDI Main Output turned on. Is there a specific Aitum setting needed or a bug report to reference?

Also, what specifically in the OBS Output API is failing in Aitum Stream Suite when the NDI Main Output is created first?

exeldro:
In Aitum Stream Suite have an extra canvas add a source to that canvas so you can see that gets rendered. Enable the main output in DistroAV. Restart OBS and see the extra canvas does not get rendered anymore.

OBS does not allow updating canvases when an output is active. The call that fails is `obs_canvas_reset_video`

BitRate27:
I was able to duplicate the issue, no source shown in "extra canvas". After incorporating this PR into the DistroAV plugin, I can see the source displayed in the "extra canvas" on startup. I also, turned NDI Main and Preview off/on, started OBS with and without outputs on, and everything continued to work as before. 

This PR seems to workaround a startup sequence issue with plugins in OBS, perhaps specific to the new Canvas API, by delaying the creation of OBS output until later. This workaround should be documented as other plugins which support output will also need to change. Another option is to provide a new event (eg. OBS_FRONTEND_EVENT_CANVASES_CREATED) to be a more friendly with Aitum type plugins and fit in better with the current startup sequence API.

exeldro:
This PR works around the issue that DistroAV starts outputs on startup, not around an issue in OBS itself.
The starting of outputs at startup blocks users from changing any video setting.

PR #1401, opened 2026-02-13
Add sync timestamp regression tests and fix deployment targets

zbynekdrlik:
## Summary
- Add 54-assertion unit test suite (`src/tests/test-sync-timestamps.cpp`) covering all timestamp math that caused recent bugs: double-conversion, data race, render timestamp, delay chain consistency, frame matching, and timecode translation
- Add post-deploy SYNC log regression test (`.github/scripts/sync-log-regression-test.sh`) that validates live telemetry consistency after deployment
- Fix push.yaml deploy target from `stream.lan` to `strih.lan` for master/main branch deployments
- Add checkout step + SYNC regression test step to both dev-ci-cd.yaml and push.yaml deploy jobs

## Test plan
- [x] Unit tests compile with `g++ -std=c++17 -Wall -Wextra` — 54/54 pass
- [ ] CI pipeline builds and format checks pass
- [ ] Dev deploy triggers SYNC log regression test against stream.lan
- [ ] Master deploy targets strih.lan (not stream.lan)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/scripts/sync-log-regression-test.sh
+++ b/.github/scripts/sync-log-regression-test.sh
@@ -1,0 +1,156 @@
+#!/bin/bash
+# SYNC Log Regression Test
+# Validates internal consistency of DistroAV SYNC log lines from stream.lan
+#
+# Checks:
+#   - creation + net == receive (timestamps match delays)
+#   - total == net + rel + obs (delays sum correctly)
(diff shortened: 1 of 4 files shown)


zbynekdrlik:
Opened on wrong repo — moving to fork

Issue #1403, opened 2026-02-17
[Bug]: Frame rate of OBS drops when using Keep Active and multiple scenes with 4K 60fps sources


BitRate27:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exist on the latest available version of DistroAV
- [x] I have checked that my OS / OBS / NDI versions are supported

### OBS Log [URL] - Important!

https://obsproject.com/logs/pJ6rb8vta0LteEHf

### Issue reported - Current Behavior

As reported by user neoinnovation in Discord:
```
I have an issue where I'm using 4 cameras running HX3 at 4K60.  My system can easily run all 4 in 4 separate OBS instances each with their own camera or with NDI Monitor. However when loading multiple cameras into 1 OBS instance it will stutter even at 2 cameras and 4 is completely unusable. Any idea on how to get it running in 1 OBS instance?
It also runs fine on my M1 Max MacBook Pro. It is my windows 11 computers that have issues.
```

### Expected Behavior

OBS performance should not be affected by multiple NDI sources not being shown.

### Steps To Reproduce

To duplicate:

Create 4 scenes with a 4K 60fps NDI source in each.
Change behavior in each source to be "Keep Active".
Start all 4 sources external to OBS so they are sending NDI data.

OBS Will become sluggish and fps rate drops to 1fps

### Environment

```markdown
- OS: Windows 11
- OBS: 32.0.2
- DistroAV: 6.1.1
- NDI: 6.3.0.3
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [ ] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)

BitRate27:
More from the user reporting the problem:
Everything is 3840x2160 60Hz. Canvas, cameras, encode.

My scenes are:

Scene: Start
Browser Source: Streamline
NDI Source: Start Cam

Scene: Pool 1
Browser Source: Streamline
NDI Source: Pool 1 Cam
NDI Source: Start Cam

Scene: Pool 2
Browser Source: Streamline
NDI Source: Pool 2 Cam
NDI Source: Start Cam

Scene: Turn
Browser Source: Streamline
NDI Source: Turn Cam
[...]

PR #1404, opened 2026-02-17
Don't receive frames from NDI if not showing

BitRate27:
This is to fix issue [1403](https://github.com/DistroAV/DistroAV/issues/1403) [Bug]: Frame rate of OBS drops when using Keep Active and multiple scenes with 4K 60fps sources.

When Keep Active is chosen, the source receiving thread keeps running whether or not the source is actually showing anywhere in OBS. This effects the performance of OBS by wasting time receiving frames from NDI and passing them to OBS (which ends up ignoring). 

The fix is simply to continue to the top of the NDI source thread loop before receiving frames if the source is not being shown.

Verified the fix works by creating 4 4K 60fps NDI sources loading a scene collection of 4 scenes with a NDI source in each. I was able to switch between scenes in Normal mode with no delay in seeing the feed. Also tested in Studio mode with 2 scenes shown and witnessed a drop in frame rate from 30 to 7.  When displaying a Multiview OBS becomes unusable, fps < 0.4.


Trouffman:
@BitRate27 thanks for the fix.

"Also tested in Studio mode with 2 scenes shown and witnessed a drop in frame rate from 30 to 7. When displaying a Multiview OBS becomes unusable, fps < 0.4."

Does that happen WITH the fix or that is the behavior without the fix ?

BitRate27:
> @BitRate27 thanks for the fix.
> 
> "Also tested in Studio mode with 2 scenes shown and witnessed a drop in frame rate from 30 to 7. When displaying a Multiview OBS becomes unusable, fps < 0.4."
> 
> Does that happen WITH the fix or that is the behavior without the fix ?

This is with the fix! It only really helps if in Normal mode. Studio Mode may be OK on a better machine, but Multiview will still suffer as all the sources are considered "showing", which is what it does now without the fix. I'll also add I am receiving a Full NDI feed generated locally, so HX3 may not have the same experience.

I am considering another idea I had a while back and that is adding a Bandwidth option called "Low when showing but not on program".  When chosen, it will use the low bandwidth setting on the receiver if it is showing, but not active. Would need to prototype it as there may be some latency as a new receiver needs to be created when switching from non active to active.

Trouffman:
Great! Thanks. That improvement can be logged in and managed in another dev cycle

PR #1405, opened 2026-02-17
Only show preview tally if source is shown and not active

BitRate27:
This fixes a long standing issue with tally, first reported in issue: https://github.com/DistroAV/DistroAV/issues/687

This problem is occurring because when a source is in Program, it is both active and showing. This causes the preview and program tally lights to be set. 

To fix this using the existing OBS API, we will only show the preview tally if the source is showing and not active. This means DistroAV will never set on_preview and on_program to true. It is only one or the other, with on_program taking priority. These are the two scenarios which change...

1. Here is when the source is on program, no green line (on_preview) for the cyan scene, only the red line (on_program) is shown.
ProgramNotGreen

2. Here is when the source is on program AND on preview, but we only show the red line (on_program) as it takes priority.
ProgramOnlyInBoth

To handle the more robust case of number 2, a change is needed in the OBS API to identify when a source is actually on preview.


Trouffman (approved):

src/ndi-source.cpp
```diff
@@ -1065,7 +1065,8 @@ void ndi_source_update(void *data, obs_data_t *settings)
 
 	// Update tally status
 	auto config = Config::Current();
-	s->config.tally.on_preview = config->TallyPreviewEnabled && obs_source_showing(obs_source);
+	s->config.tally.on_preview = config->TallyPreviewEnabled && obs_source_showing(obs_source) &&
```
paulpv: Please refactor this to pull out the compound on_preview/on_program logic into dedicated helper functions (`is_on_preview()` `is_on_program()`?) and call them as appropriate.

paulpv (changes requested):

src/ndi-source.cpp
```diff
@@ -884,6 +884,16 @@ int safe_strcmp(const char *str1, const char *str2)
 	return strcmp(str1, str2);
 }
 
+bool tally_on_preview(obs_source_t *source)
```
paulpv, line 887: I love it!

src/ndi-source.cpp
```diff
@@ -884,6 +884,16 @@ int safe_strcmp(const char *str1, const char *str2)
...
+{
+	return (Config::Current())->TallyPreviewEnabled && obs_source_showing(source) && !obs_source_active(source);
+}
+
+bool tally_on_program(obs_source_t *source)
```
paulpv, line 892: I love it

paulpv (approved):

PR #1409, opened 2026-02-19
Change to metadata only if not showing

BitRate27:
This is to fix issue 1403 [Bug]: Frame rate of OBS drops when using Keep Active and multiple scenes with 4K 60fps sources.

When Keep Active is chosen, the source receiving thread keeps running whether or not the source is actually showing anywhere in OBS. This effects the performance of OBS by wasting time receiving frames from NDI and passing them to OBS (which ends up ignoring).

One fix is simply to continue to the top of the NDI source thread loop before receiving frames if the source is not being shown.

The other change is to change the receiver bandwidth to metadata only so NDI does not process the video data for frames that are not going to be shown. When the NDI feed is shown again, then the receiver is recreated. This should be faster as it is done in sequence and not after a period of time.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -397,6 +397,11 @@ void *ndi_source_thread(void *data)
 	NDIlib_recv_create_v3_t recv_desc;
 	recv_desc.allow_video_fields = true;
 
+	NDIlib_recv_create_v3_t meta_recv_desc{};
+	meta_recv_desc.bandwidth = NDIlib_recv_bandwidth_metadata_only; // no decode cost
+	meta_recv_desc.allow_video_fields = false;
+	bool last_showing = true;
+
 	NDIlib_recv_instance_t ndi_receiver = nullptr;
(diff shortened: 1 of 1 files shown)


paulpv (commented):
What if you just created both receivers at the beginning and then more simply just toggled between the two without having to destroy one and then create the other. 
Would that work?

BitRate27:
The user confirmed the problem stopped when we changed to a metadata only receiver. The change to using the metadata shows the problem lies in NDI's handling of multiple receivers in the same library instance. Looks like the delay when switching from metadata to A/V is the same or maybe even worse than pausing/resetting the connection when not showing (Not Keep Alive). Another thing to try is initializing a full instance of the NDI library for every source. This would mean only one receiver per instance of the library in the exe. I could try it out, but with the small delay I am seeing without the Keep Alive (basically the existing code) it seems hard to justify the change.

BitRate27:
Closing this pull request as it doesn't work any better than pausing/resetting when not shown.

PR #1420, opened 2026-02-28
Use separate process (ndi-server.exe) for each NDI receiver

BitRate27:
This is to fix issue 1403 [Bug]: Frame rate of OBS drops when using Keep Active and multiple scenes with 4K 60fps sources.

This change implements a ndi-server.exe which runs in a separate process with only one NDI receiver in the process.

DistroAV launches an ndi-server.exe for every source thread and OBS process, so you can run multiple OBSs.

This is only for Windows during the proof-of-concept. The ndi-server.exe needs to be in the same folder as the distroav.dll. The installation.exe will install the ndi-server.exe in the obs-plugins folder. If the exe is not in the plugins folder, the plugin will behave as it does before this change.

You can tell the server is running by the multiple console windows that will appear, one for every NDI source. 

The ndi-server.exes will be killed when OBS exits.

This PR is a proof of concept, to be tested by various users before deciding to include.


BitRate27:
Closing this as further tests showed no improvement in performance.

Issue #1449, opened 2026-04-22
[Bug]: Output Filter on a source not starting when using 6.2.0


Trouffman:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exists on the latest available version of DistroAV.
- [x] I have checked that my OS / OBS / NDI versions are supported & up-to-date.

### Issue reported - Current Behavior

With 6.2.0 if the NDI Output filter is on a scene, it will not auto-start when launching OBS

### Expected Behavior

Filter automatically start when OBS starts

### Steps To Reproduce

1. Launch OBS
2. Add a source + a filter on the Source
3. close OBS
4. Launch OBS
5. Filter on the source is not started.

### Environment

```markdown
DistroAV 6.2.0
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [x] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)
- [x] I am filing this bug report as a Human and have reviewed its content to the best of my ability to ensure it's accurate.

BitRate27:
After further investigation by @paulpv, it was determined this occurs with audio only filters. This is fixed in [PR 1452](https://github.com/DistroAV/DistroAV/pull/1452) and should be included in 6.2.1.

PR #1452, opened 2026-04-23
Add filter_add so all filter sender are created at startup

BitRate27:
This is to fix (https://github.com/DistroAV/DistroAV/issues/1449). 
The regression was caused by "Default NDI name for filters to be `<filter> (<source>)`" (https://github.com/DistroAV/DistroAV/pull/1412)

Instead of checking to see if the sender needs creating in the filter_tick (which is not called on an audio only filter) we now do  that check in the new ndi_filter_add function. 

A side note on implementation: copilot suggested numerous changes to fix this, one of them being to add a filter_add callback. The other changes suggested by copilot were to supposedly fix a race condition which could not be duplicated so those suggested changes were not included in this PR.

***This is a regression and should be included in a hot fix in 6.2.1***

### DIFF
--- a/src/ndi-filter.cpp
+++ b/src/ndi-filter.cpp
@@ -530,15 +530,16 @@ void ndi_filter_tick(void *data, float)
 {
 	auto f = (ndi_filter_t *)data;
 
-	if (!f->ndi_sender) {
-		obs_source_t *parent = obs_filter_get_parent(f->obs_source);
-		if (parent) {
-			ndi_sender_create(f, nullptr);
-		}
-	}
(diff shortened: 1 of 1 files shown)


Trouffman:
It solve the problem for the Audio Filter.

BUT the audio is "always running" as well (aka not dependent on the parent / source being active/showing or not.)

Trouffman:
Tested and confirmed to work on both audio & video filter.

Trouffman:
This will be merged. Acknowledging the research work by @paulpv here :)

Trouffman (approved):
Tested and address the issue it fixes.

PR #1464, opened 2026-06-13
Fix NDI frame resize mid stream from Teams

### LINKED ISSUE #1463: [Bug]: NDI Source changes scaling randomly when using NDI from Teams
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exists on the latest available version of DistroAV.
- [x] I have checked that my OS / OBS / NDI versions are supported & up-to-date.

### OBS Log [URL] - Important!

_No response_

### Issue reported - Current Behavior

Hello, I am experiencing intermittent scaling/zoom glitches on Microsoft Teams NDI sources in OBS.

During live productions with 4 Teams participants, some NDI sources briefly change size and/or zoom level for a fraction of a second before returning to normal. The issue can affect one or multiple sources and may occur several times during a live session.

The same NDI feeds remain perfectly stable in NDI Video Monitor while the corresponding sources glitch in OBS.

I have already tested with a simplified OBS project, reinstalled OBS, DistroAV and the NDI Runtime, and toggled Framesync and Hardware Acceleration without success.

- NDI feeds remain perfectly stable in NDI Video Monitor.
- In OBS, some NDI sources briefly change scale/zoom multiple times during a live session.
- The issue can affect a single source while all four NDI sources are visible simultaneously.
- OBS, DistroAV and NDI Runtime have all been completely reinstalled.
- Framesync ON/OFF: no effect.
- Hardware Acceleration ON/OFF: no effect.
- Simplified OBS scene collection (4 NDI sources + background only): issue still occurs.

Thank you for any advices/information, I've been looking for a solution for days now.

### Expected Behavior

- NDI feeds remain perfectly stable in NDI Video Monitor.
- In OBS, some NDI sources briefly change scale/zoom multiple times during a live session.
- The issue can affect a single source while all four NDI sources are visible simultaneously.
- OBS, DistroAV and NDI Runtime have all been completely reinstalled.
- Framesync ON/OFF: no effect.
- Hardware Acceleration ON/OFF: no effect.
- Simplified OBS scene collection (4 NDI sources + background only): issue still occurs.

### Steps To Reproduce

Additional observations:

- NDI Video Monitor remains stable while the corresponding source in OBS glitches.
- The issue appears more frequently shortly after switching to a new Teams meeting and reassigning existing DistroAV sources to new Teams NDI feeds.
- During one live session, the NDI sources were configured about 20 minutes before going live and only a few glitches occurred.
- During another live session, the sources were reassigned about 30 seconds before going live, and the glitches occurred much more frequently.

### Environment

```markdown
macOS Tahoe
OBS Studio 32.1.2
DistroAV 6.2.1
NDI Runtime 6.3.2
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [x] I have confirmed this bug exist on more than one device/computer
- [x] I have tested using a new scene collection to reproduce the issue
- [x] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)
- [x] I am filing this bug report as a Human and have reviewed its content to the best of my ability to ensure it's accurate.

BitRate27:
This PR is to fix https://github.com/DistroAV/DistroAV/issues/1463 [Bug]: NDI Source changes scaling randomly when using NDI from Teams. 

There appears to have been recent changes to how Teams sends NDI sources, mainly an altering of the width and height mid stream for various reasons.

To fix the scaling issue, the user must set the Transform of the NDI source to Fit to Screen (Ctrl-F), or Edit the Transform (Ctrl-E) and set Bounds to "Cover". 

This will keep the scaling of the image the same in OBS as the Teams NDI source changes. However, there is flashing of the image in OBS where it sometimes momentarily shows a smaller version in the Preview/Program windows. 

A solution to the flashing problem was suggested in [the issue](https://github.com/DistroAV/DistroAV/issues/1463). Because we are sending async frames, the frame size returned by get_width and get_height will be ahead of what is about to be rendered. The suggested solution is to remove these callbacks from the ndi_source_info so the async sizes will be used instead.

To test this change, I created a test NDI source which alters the frame size between 1600 and 800 wide, every second.
I was able to duplicate the scaling issue. After applying the proper Transform to the source the scaling issue went away, but the flashing persisted. After removing the get_width and get_height callbacks, the flashing went away. I also tested by creating new NDI sources and using old NDI sources and everything worked fine. Also confirmed the empty frame behavior still works when receiving NDI output from OBS.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -1230,18 +1230,6 @@ void ndi_source_destroy(void *data)
 	obs_log(LOG_DEBUG, "'%s' -ndi_source_destroy(…)", obs_source_name);
 }
 
-uint32_t ndi_source_get_width(void *data)
-{
-	auto s = (ndi_source_t *)data;
-	return s->width;
-}
-
-uint32_t ndi_source_get_height(void *data)
-{
-	auto s = (ndi_source_t *)data;
-	return s->height;
-}
-
 obs_source_info create_ndi_source_info()
 {
 	// https://docs.obsproject.com/reference-sources#source-definition-structure-obs-source-info
@@ -1262,9 +1250,6 @@ obs_source_info create_ndi_source_info()
 	ndi_source_info.hide = ndi_source_hidden;
 	ndi_source_info.deactivate = ndi_source_deactivated;
 	ndi_source_info.destroy = ndi_source_destroy;
-
-	ndi_source_info.get_width = ndi_source_get_width;
-	ndi_source_info.get_height = ndi_source_get_height;
 
 	return ndi_source_info;
 }


Trouffman:
Great work on the testing and testing procedure.
Review in progress.

Trouffman (approved):

Issue #1438, opened 2026-04-21
[Bug]: Output Name not updated when changed via Websocket


Trouffman:
### Is there an existing issue for this?

- [x] I have searched the existing issues and did not find a similar issue

### This bug exists on the latest available release of DistroAV and supported environment?

- [x] I have checked this bug still exists on the latest available version of DistroAV.
- [x] I have checked that my OS / OBS / NDI versions are supported & up-to-date.

### Issue reported - Current Behavior

When updating the Outpu Name via Websocket, it is not updated in the Settings Output window.

<img width="574" height="259" alt="Image" src="https://github.com/user-attachments/assets/ebd85d2b-4457-4f26-9f1e-cb834ad5bcf6" />

Initially reported by Warchamp7 (OBS) on a discord discussion.

### Expected Behavior

Get the active name set.

### Steps To Reproduce

-

### Environment

```markdown
- 6.2.0 as well
```

### Pre-Check for Bug Report

- [x] I am able to reproduce the bug following the steps provided above
- [ ] I have confirmed this bug exist on more than one device/computer
- [ ] I have tested using a new scene collection to reproduce the issue
- [ ] I have tested on a 'clean' OBS installation (aka with no other scripts or plugin installed)
- [ ] I am filing this bug report as a Human and have reviewed its content to the best of my ability to ensure it's accurate.

BitRate27:
If the user is trying to change the ndi_name in the Output settings, then I created a html page to test this bug: 

[obs_output_settings_modify.html](https://github.com/user-attachments/files/29109791/obs_output_settings_modify.html)

Download the html file and open it in a web browser. 

The html file will return a list of Output names, selecting one of the Output Names will show a list of settings associated with that Output. You can then modify the field to duplicate the issue reported. 

FYI: The Output Name that is returned using the WebSocket interface is a fixed name, hardcoded, that is not exposed to the user.
DistroAV creates a preview output with a name of NDI Preview Output, and creates a main output with the name of NDI Main Output. It is not clear if the user is talking about Output Name or ndi_name in settings.

@Trouffman  How did WarChamp change the Output Name using WebSockets?

PR #1469, opened 2026-06-19
Restart output if output settings are updated

BitRate27:
This PR is to fix issue https://github.com/DistroAV/DistroAV/issues/1438 [Bug]: Output Name not updated when changed via Websocket

To fix the issue, the ndi_output_update callback was modified to:
1. Check if the output is running, if so, stop it.
2. Save the changes to the config depending on what output was updated.
3. Restart the output if it was running.

This fixes the issue reported because the ndi_output_update callback is called when the settings are changed by the websocket API.

Here is a web page used to test this PR. The web page will allow the user to change the output settings via the websocket API: 
[obs_output_settings_modify.html](https://github.com/user-attachments/files/29147249/obs_output_settings_modify.html)

To test this PR:
1. Download the html file.
2. Open the html file in a web browser
3. Launch OBS
4. Select Connect on web page
5. Select NDI Main Output or NDI Preview Output
6. Change the name and/or group and select Apply
7. See the NDI name change in NDI Studio Monitor
8. Open DistroAV NDI Settings dialog
9. Verify name and/or group changed


src/ndi-output.cpp
```diff
@@ -254,18 +255,55 @@ bool ndi_output_start(void *data)
...
+		ndi_output_stop(o, 0);
+	}
+
 	o->ndi_name = name;
 	o->ndi_groups = groups;
 	o->uses_video = obs_data_get_bool(settings, "uses_video");
 	o->uses_audio = obs_data_get_bool(settings, "uses_audio");
 
+	auto outputName = obs_output_get_name(o->output);
+	if (outputName && std::strcmp(outputName, "NDI Main Output") == 0) {
```
Trouffman, line 282: Do we need this whole hard-coded logic here, instead of the defined logic for start/stop, including the recent change that "test pixel format compatibility & naming compatibility"

BitRate27: Fixed in consolidated PR https://github.com/DistroAV/DistroAV/pull/1472

BitRate27:
Closing this PR as this is all handled in a bigger PR https://github.com/DistroAV/DistroAV/pull/1472

PR #1470, opened 2026-06-23
Simplify check for supported pixel format

BitRate27:
This PR simplifies the test for pixel format supported. The old test created a full NDI Main Output, including a NDI sender just to check if the current pixel format (OBS->Settings->Advanced) is supported by NDI. This created the following lines in the logs file without verbose or debug turned on:

```
14:02:58.238: [distroav] NDI Output Updated. 'NDI Output Support Test 49585'
14:02:58.241: [distroav] NDI Output started successfully. 'NDI Output Support Test 49585'
14:02:58.241: Output 'NDI Main Output': stopping
14:02:58.241: Output 'NDI Main Output': Total frames output: 0
14:02:58.241: Output 'NDI Main Output': Total drawn frames: 0
14:02:58.242: [distroav] NDI Output Stopped. 'NDI Output Support Test 49585'
```

These messages can be misleading, as it appears that the NDI Main Output is stopping. Actually, only the Support Test output was stopped.

To simplify, a new output is registered with OBS, called "test_ndi_output". This output will only test if the video format supported in the test_output_start callback. The main_output_is_supported() method was modified to use this new test_ndi_output instead of creating a Main Output. 

Simplifying this also makes it easier to fix issues related to web-sockets changing output settings.

Tested by turning off output, selecting invalid pixel format in Settings->Advanced, bringing up the DistroAV NDI Settings dialog and witnessing the error code and that the UX is the same as before. Also tested by loading profile with a bad pixel format, and a good pixel format and witnessed the output starting and stopping and the Output Settings UI reflecting the current state.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -81,6 +81,7 @@ target_sources(
     src/ndi-finder.h
     src/ndi-finder.cpp
     src/ndi-output.cpp
+    src/test-output.cpp
     src/ndi-source.cpp
     src/plugin-main.cpp
     src/plugin-main.h
--- a/src/main-output.cpp
+++ b/src/main-output.cpp
@@ -102,33 +102,13 @@ void main_output_start()
 bool main_output_is_supported()
 {
 	obs_log(LOG_DEBUG, "+main_output_is_supported()");
-	auto config = Config::Current();
-	auto output_name = config->OutputName;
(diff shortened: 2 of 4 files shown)


Trouffman (approved):
Thank you for simplifying this.
The output test could also be extend further if we need to test for more than just the pixel support.

PR #1477, opened 2026-07-18
Source: Fix alpha transparency in Low/Lowest

### LINKED ISSUE #937: [Bug]: NDI transparency does not work when using low/lowest latency
### Operating System Version

MacOS 13.5

### OBS Version

30.0.0

### NDI Tools Version

5.6

### Describe the bug

Since upgrading to the latest NDI drivers and OBS release. I can no longer see any transparency on slides sent from ProPresnter 7. Objects that should be transparent are completely opaque. I've double-checked that the alpha channel is turned on on both ends, and have tested it from multiple different machines, both sending and receiving the signal. I've also tested receiving the NDI signal in Newtek's NDI virtual monitor, and that seems to show the transparency fine, so it seems to be a problem with the OBS plugin. 

### Steps to reproduce

_No response_

### Expected behavior

_No response_

### Screenshots

_No response_

### Additional context

_No response_

scr512:
> **Disclosure:** this fix was diagnosed, implemented and tested with the help of an AI coding agent (LLM). All results were reproduced and reviewed by a human before submission.

Fixes #937.

**Problem:** NDI sources carrying alpha (ProPresenter, FreeShow, vTube Studio, …) lose transparency when the source latency mode is Low or Lowest.

**Root cause:** with `NDIlib_recv_color_format_fastest`, the NDI runtime decodes alpha-carrying streams to **UYVA** (UYVY + separate alpha plane). `ndi_source_thread_process_video2()` maps UYVA → `VIDEO_FORMAT_UYVY`, silently discarding the alpha plane. Verified with a local probe receiver against a BGRA test source: `UYVY_BGRA` → BGRA with alpha intact; `fastest` → UYVA.

**Fix:** request `NDIlib_recv_color_format_UYVY_BGRA` for all latency modes. Non-alpha sources still decode zero-copy to UYVY (same as `fastest`); alpha sources are converted to BGRA by the NDI runtime.

**Notes:**
- #954 previously fixed this for Low, but it only ever landed on `develop` and was lost in the 6.0.0 rewrite — no shipped release contains it. This PR extends the same approach to Lowest as well.
- The "proper" fix (native UYVA support, #1261) requires `VIDEO_FORMAT_UYVA` in libobs (obsproject/obs-studio#10345, closed unmerged). This is a minimal alternative that works with stock OBS today.
- Could a maintainer please add the `Seeking Testers` label so CI packages test installers?

**Testing:** built for macOS (6.2.1) and Windows x64 (this branch); verified alpha checkerboard and real ProPresenter/FreeShow feeds render with transparency in **Normal, Low, and Lowest**. Plugin loads cleanly; no new warnings.

### DIFF
--- a/src/ndi-source.cpp
+++ b/src/ndi-source.cpp
@@ -462,10 +462,12 @@ void *ndi_source_thread(void *data)
 			//
 			// Update recv_desc.latency
 			//
-			if (s->config.latency == PROP_LATENCY_NORMAL)
-				recv_desc.color_format = NDIlib_recv_color_format_UYVY_BGRA;
-			else
-				recv_desc.color_format = NDIlib_recv_color_format_fastest;
+			// "fastest" decodes alpha-carrying sources to UYVA,
+			// whose separate alpha plane is then dropped when
+			// mapped to VIDEO_FORMAT_UYVY in process_video2().
+			// Use UYVY_BGRA for all latency modes: non-alpha
+			// sources still decode to UYVY (zero-copy). #937
+			recv_desc.color_format = NDIlib_recv_color_format_UYVY_BGRA;
 			obs_log(LOG_DEBUG,
 				"'%s' ndi_source_thread: reset_ndi_receiver; Setting recv_desc.color_format=%d",
 				obs_source_name, //


Lordmau5:
Gave this a test in my local developer environment.

I feel there might've been some changes in one of the recent NDI SDK builds that seems to be either fixing it or improving performance for converting to BGRA and sending. That, *or* it still uses higher bandwidth than UYVA, unsure.

I don't have time to thoroughly test it for those differences but on a quick glance it works and I don't seem to be noticing increased latency even in the low and lowest modes (though it does feel like low is very similar to lowest, I don't remember what else the difference was under the hood in DistroAV's code)

BitRate27:
I was able to duplicate the problem with UYVA. This does indeed fix the issue. The only problem is OBS obsoleted the obs_properties_add_button, which is causing the Linux build to fail. I will make a PR to fix that (call obs_properties_add_button2 instead) and then merge that into master before approving this and merging it to master.

Lordmau5:
Please don't merge this yet - I've had the chance to test it a bit more thoroughly and while it at first looked like it worked fine, in my stream yesterday when using it to send a source from one of my PCs to the other (both windows 11) I had massive lag in the NDI source and the only fix was to change it to Lowest Bandwidth mode (which in turn lowers the quality by quite a bit)

I'll switch back to the latest public release for my stream today and see if UYVY has better performance again.

PR #1487, opened 2026-08-26
Network Monitor - Report status of Network Adapters, NDI Senders and NDI Receivers in real time

BitRate27:
## Add Network Monitor: live diagnostics for NDI adapters, senders, and receivers

Adds a Network Monitor dialog (reachable from Output Settings) with three live-updating tabs to help diagnose why NDI discovery or streaming isn't working, instead of guessing at OBS Studio's own logs.

### Adapters tab
- Lists every network interface with IPv4/mask, link state, multicast capability, virtual-adapter detection, and live throughput/link speed.
- **Windows**: per-adapter Windows Firewall diagnostics — network category (Public/Private/Domain), firewall on/off, and whether mDNS (5353), NDI's TCP ports (5959-6000), and File & Printer Sharing are actually allowed through, evaluated per adapter rather than one machine-wide answer. This matters because a machine can have adapters on different network profiles simultaneously, and a single aggregate check can silently misreport the one carrying NDI traffic.
- Live mDNS multicast join + round-trip tests per adapter (catches VPN clients that block LAN multicast at the driver level even when local socket checks pass).
- Linux: interface enumeration plus Avahi daemon detection (no OS-bundled mDNS responder on Linux).

### Senders / Receivers tabs
- Per-source and per-receiver live stats: fps, timestamp-based fps, jitter, sync drift, send-blocking %, processing %, and connection/discoverability changes over time.

### Shared UI
- Sortable, reorderable, show/hide-able columns (right-click header), persisted layout.
- Copy-to-clipboard as a monospace ASCII table; "dump to OBS log" for bug reports; help button linking to the wiki.

### Supporting changes
- `ChangeNotifier` mechanism so background-thread stat updates push to the UI live instead of on a fixed poll.
- NDI source properties gained an inline diagnostics report.
- Output Settings: apply-on-change, keeps outputs running across settings changes, config now also saves when changed via the OBS WebSocket API. Also real time feedback of output status using green checkmarks, or red x's.

### Not yet done / known limits
- macOS firewall/port checks are not implemented (documented as a platform limitation).
- File & Printer Sharing / mDNS / NDI-port checks only recognize rules scoped to this process's own executable, or (for File & Printer Sharing) Windows' built-in unscoped rule group — a custom rule under a different name/program won't be picked up.


BitRate27:
Items found while testing not yet fixed:
1. Config tab not output to OBS log.
2. Config tab has no Copy button.
3. Put up confirmation dialog when Dump to OBS Log is selected. Show how to access the log within OBS.
4. Keep user selected column width, don't revert back to max size.
5. Allow resize of Network Monitor to a smaller size. Add scroll bars if too small.
6. Save sort column/direction, width, order, and show/hide for all three tables in user config. (After testing complete)

BitRate27:
Added copy and dump of config tab, fix random large jitter ratio.
Allow resize of dialog, move copy to bottom of dialog and fix ascii text.
Store network monitor position in user config.

